PR #11, opened 2019-03-12
Add support for running examples via cargo

### LINKED ISSUE #10: Add a way to run cargo examples
Currently, only binary targets seem to be supported.

killercup:
Fixes #10

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,8 +34,12 @@ struct Opt {
     release: bool,
 
     /// Binary to run
-    #[structopt(short = "b", long = "bin")]
+    #[structopt(short = "b", long = "bin", conflicts_with = "example")]
     bin: Option<String>,
+
+    /// Example to run
+    #[structopt(long = "example", conflicts_with = "bin")]
+    example: Option<String>,
 
     /// Other command to run
     #[structopt(short = "e", long = "exec")]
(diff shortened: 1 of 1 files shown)


jonhoo:
I wonder -- should we make `bin`, `example`, [and `test`](https://www.reddit.com/r/rust/comments/b08oex/cargoflamegraph_generate_flame_graphs_directly/eidp5e3/) be subcommands instead of just flags? That way they could also have their own flags.

killercup:
Hm, I thought the arguments were meant to be similar to cargo-run -- but that doesn't deal with tests or bechs so I'm not sure.

jonhoo:
That's what I mean though. Maybe it should be:
```console
$ cargo flamegraph run --bin foobar
$ cargo flamegraph run --example foobar
$ cargo flamegraph test
$ cargo flamegraph bench
```

robinst:
I was just looking if there was an open issue for running flamegraph for tests, being able to do `cargo flamegraph test` or `cargo flamegraph test mytestcase` would be nice!

spacejam:
Thanks @killercup :D

killercup:
I probably won't have time to change this PR for a while -- feel free to submit code based on it or merge as-is :)

spacejam:
@jonhoo hmmm that does seem reasonable and consistent. I would like to default to the run behavior if no other command is entered, so that you can just type `cargo flamegraph` and have it generate what you need, but I think that's an approach that scales to others

spacejam:
@killercup NP I fixed the conflict!

PR #13, opened 2019-03-12
Fix dtrace implementation

Licenser:
With -c we should `pid == $target` instead of trying to match the
executable name. This is more performant and more correct as it
will only match exactly this process not any process with this
name.


spacejam:
Nice reduction of complexity, thanks!

PR #17, opened 2019-03-13
build with --release by default

matklad:
Also check if debug = true is enabled in Cargo.toml by parsing output

closes https://github.com/ferrous-systems/cargo-flamegraph/issues/15

This takes the simplest possible approach of fishing for `debuginfo` in 

```
   Compiling crt v0.1.0 (/home/matklad/projects/crt)
    Finished release [optimized + debuginfo] target(s) in 8.94s
```

The *proper* way to handle this would be to run the build with `--message-format=json` and parsing it, which is not too complicated, but still seems unnecessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,12 @@ like `~/.cargo/bin`.
 ## Examples
 
 ```
-# defaults to profiling cargo run, which will
-# also profile the cargo compilation process
-# unless you've previously issued `cargo build`
+# defaults to profiling cargo run
 cargo flamegraph
 
-# if you'd like to profile your release build:
-cargo flamegraph --release
+# by default, `--release` profile is used, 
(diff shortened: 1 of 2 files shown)


matklad:
Once https://github.com/oli-obk/cargo_metadata/pull/72 is merged, we might switch to *proper* way to find the executable in the `target` dir.

spacejam:
FYI I just updated the logic around debuginfo detection to rely on cargo_metadata parsing some info. I was getting the warning whether I had `debug = true` or not before.

https://github.com/ferrous-systems/cargo-flamegraph/commit/a802af1baddeda9d5c3c6203c0d1785bd81beb24#diff-639fbc4ef05b315af92b4d836c31b023R200

PR #26, opened 2019-03-14
SIGINT handling

ibabushkin:
In some cases it helps to sample a longrunning process for some time and interrupt the whole procedure after a reasonably short period of time.

This can be done by simply terminating the traced process, but a transparent handling of SIGTERM is certainly easier to use. This commit just registers an empty handler and checks that the return value of the child indicates a termination by signal or success.


src/lib.rs
```diff
@@ -117,18 +122,41 @@ pub fn generate_flamegraph_by_running_command<
     workload: String,
     flamegraph_filename: P,
 ) {
+    eprintln!("{:?}", flamegraph_filename);
```
jonhoo: You probably didn't intend to include this :)

src/lib.rs
```diff
@@ -117,18 +122,41 @@ pub fn generate_flamegraph_by_running_command<
...
 
+    #[cfg(not(unix))]
+    {
+        if !exit_status.success() {
+            eprintln!("failed to sample program");
+            std::process::exit(1);
+        }
+    }
+
+    eprintln!("perf exited, so now we wait: {:?}", flamegraph_filename);
```
jonhoo: ditto

ibabushkin:
On 2019-03-14, Jon Gjengset wrote:
> jonhoo commented on this pull request.
> 
> 
> 
> >      }
>  
> +    #[cfg(not(unix))]
> +    {
> +        if !exit_status.success() {
> +            eprintln!("failed to sample program");
> +            std::process::exit(1);
> +        }
> +    }
> +
> +    eprintln!("perf exited, so now we wait: {:?}", flamegraph_filename);
> 
> ditto
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/ferrous-systems/flamegraph/pull/26#pullrequestreview-214788577

Ah dang, forgot that :)

src/lib.rs
```diff
@@ -117,16 +122,35 @@ pub fn generate_flamegraph_by_running_command<
     workload: String,
     flamegraph_filename: P,
 ) {
+    // handle SIGINT by doing nothing
```
jonhoo: I'll let @spacejam comment on this, but personally I think I'd like to see a comment here explaining why we install an empty signal handler. Just to aid the curious reader who happens to come across this code a long time for now :)

ibabushkin: Indeed, something more useful would help. What about this bit based on the commit message?

"Handle SIGINT with an empty handler, causing only the tracing child to terminate upon delivery. This way, users can interrupt execution and still get a flamegraph from the sample data collected so far." 

Thinking about it, it could also make sense to deregister the handler as soon as the child has terminated to do the right thing when a sigint is delivered later on.

spacejam: Yeah, I like that more detailed comment!

spacejam:
Hey @ibabushkin thanks for this contribution :D To keep friction down I'll just merge as-is and add the more detailed comment myself, then I'll cut a new release that includes this.

ibabushkin:
On 2019-03-15, Tyler Neely wrote:
> Merged #26 into master.
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/ferrous-systems/flamegraph/pull/26#event-2206064092

Alright! I guess you don't care about signal handler cleanup then? It's
really minor, obviously.

spacejam:
If you'd like to add that in a separate PR I'd be glad to accept it! But low pressure from my side :]

In this case I just wanted to get it merged in because I was excited to use it xD

PR #28, opened 2019-03-15
Cleaner signal handling.

ibabushkin:
As discussed in #26, this implements a cleaner signal handling (essentially we remove the SIGINT handler as soon as the tracing child has exited in any way to allow the user to interrupt the remaining computations).


spacejam:
looks good! I'll cut a `0.1.12` release now that includes both of your changes!

ibabushkin:
Great!

Issue #27, opened 2019-03-14
Support for running tests


ibabushkin:
Given that the tool already supports running binaries and examples (and people seem to want benchmark support too), would there be interest in having the same functionality for tests? The only issue I see with this kind of functionality would be the fact that one would profile the test harness and/or multiple tests in the same binary in some cases.

Additionally, one can obviously just find the binary cargo has built and running that using `flamegraph`,  but that's somewhat annoying.

I have some hacky code for this around, but before I clean it up and open a PR I wanted to sample what opinions the authors and maintainers have.

jonhoo:
I think that's a great idea! There has already been a few requests from it (e.g., [on Reddit](https://www.reddit.com/r/rust/comments/b08oex/cargoflamegraph_generate_flame_graphs_directly/eidp5e3/)), and I proposed the syntax I would love to see [here](https://github.com/ferrous-systems/flamegraph/pull/11#issuecomment-472178617). _Any_ supported way of doing it would be awesome though :D

ibabushkin:
Alright, I'll clean up my unwrap-fest and open a PR then. Properly isolating the profiling of the actual test code and the surrounding stuff would probably require a lot of work outside the scope of this tool, so I guess it's reasonable to leave it to the user to set up something more reasonable if accurate profiling is necessary.

jonhoo:
I would probably do something like `cargo test --no-run`, and _then_ profile `cargo test`. That should eliminate most of the additional stuff!

ibabushkin:
I think there is a misunderstanding: it's rather easy to profile the running of the test only -- we just fiddle the test binary out of the `target` directory. We'd also have to ensure that the test has been built correctly, but that's easy enough. The issue remaining is that the test binary could run multiple tests (if they come from the same file, they end up in the same binary roughly speaking) and that there is some setup code in the binary as well. But working around that would be too much for now I think.

spacejam:
This is partially implemented now, but can be cleaned up a bit more.

Say I want to run a bunch of tests that begin with `pagecache_bug_` that live in a file called `test_pagecache.rs`. 

With normal cargo, I would run:
```
cargo test pagecache_bug_
```

and right now I need to run:

```
cargo flamegraph --test test_pagecache -- pagecache_bug_
```

I think we can add better detection to eliminate needing to specify the binary name, and generally match the syntax @jonhoo has suggested to bring it to the ideal situation of:

```
cargo flamegraph test pagecache_bug_
```

[...]

Issue #1, opened 2019-03-09
Add support for profiling benchmarks


bheisler:
This looks like a useful project! I've had a few folks ask for profiling support to be added to Criterion.rs; now I can direct them to cargo-flamegraph.

That being the case, it seems like it would be useful to have the ability to run `cargo bench` inside a profiler with cargo-flamegraph. Is that supported? I didn't see it in the documentation. Naturally, it should be possible to pass arguments to the benchmark executables as well, probably using the same `--` separator that Cargo itself uses.

BourgondAries:
This would be extremely useful

PR #29, opened 2019-03-15
Support for profiling tests.

ibabushkin:
This implements a `--test` parameter for `cargo-flamegraph` which finds a test target in the crate and allows to run the corresponding binary (see discussion in #27). If the code is fine in terms of functionality, it can be easily extended to support benchmarks as well (as discussed in #1).


spacejam:
w00t finally support for one of the most requested features, thank you so much @ibabushkin :D I'll cut a release shortly that includes this!

PR #33, opened 2019-04-28
Add -F option

### LINKED ISSUE #21: Passing perf record -F option
perf record can take -F option to control sampling frequency. Adding a way to pass it to perf would be useful.

Shadlock0133:
Adds -F/--freq option to cli that allows passing sampling rate to profiler.
On Linux it changes value of -F option to perf record from default 99.
On Non-Linux OSes it changes the default value off 997 passed to dtrace script.

I haven't tested these change. I only know it compiles on Linux.

Fixes #21

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -67,6 +67,10 @@ struct Opt {
     #[structopt(long = "root")]
     root: bool,
 
+    /// Sampling frequency
+    #[structopt(short = "F", long = "freq")]
+    frequency: Option<u32>,
+
     trailing_arguments: Vec<String>,
 }
 
@@ -297,5 +301,6 @@ fn main() {
         Workload::Command(workload),
         &flamegraph_filename,
         opt.root,
+        opt.frequency,
     );
 }
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -25,6 +25,10 @@ struct Opt {
     #[structopt(short = "p", long = "pid")]
     pid: Option<u32>,
 
+    /// Sampling frequency
+    #[structopt(short = "F", long = "freq")]
+    frequency: Option<u32>,
+
     trailing_arguments: Vec<String>,
 }
 
@@ -65,5 +69,6 @@ fn main() {
         workload,
         flamegraph_filename,
         opt.root,
+        opt.frequency,
     );
 }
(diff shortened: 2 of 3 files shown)


jonas-schievink:
Sorry for the long delay on this! This still seems like a reasonable addition, so it would be great if you could rebase this onto the latest master, then I'll look into merging this.

src/bin/cargo-flamegraph.rs
```diff
@@ -67,6 +67,10 @@ struct Opt {
     #[structopt(long = "root")]
     root: bool,
 
+    /// Sampling frequency
+    #[structopt(short = "F", long = "freq")]
+    frequency: Option<String>,
```
jonas-schievink: This should not accept any string here, probably something like `u16` is sufficient

src/bin/flamegraph.rs
```diff
@@ -25,6 +25,10 @@ struct Opt {
     #[structopt(short = "p", long = "pid")]
     pid: Option<u32>,
 
+    /// Sampling frequency
+    #[structopt(short = "F", long = "freq")]
+    frequency: Option<String>,
```
jonas-schievink: (same here)

src/lib.rs
```diff
@@ -45,6 +45,7 @@ mod arch {
     pub(crate) fn initial_command(
         workload: Workload,
         sudo: bool,
+        freq: Option<String>,
```
jonas-schievink: This should also accept an `Option<u16>` instead, I think

Shadlock0133:
I used String because perf allows to pass "max" as argument to -F.

jonas-schievink:
Hmm, it might be a good idea to also pass `--strict-freq` to `perf` here, since not getting the requested frequency shouldn't really silently happen.

> I used String because perf allows to pass "max" as argument to -F.

Yeah I just read the discussion in #21, and this sounds like a reasonably feature. However I feel like a `String` is still too general for this (for example, this would allow someone to splice in an arbitrary dtrace script).

I'm also worried about parity between the perf and dtrace backends here. Does dtrace have something like `max`?

If dtrace can support this as well, you can create an enum like this, and implement `FromStr` for it:

```rust
enum Freq {
    Hz(u16),
    Max,
}
```

This would ensure that only reasonably numbers or `max` are accepted, which is what we want here.

If dtrace doesn't easily support `max`, I'd prefer to not expose it to the user either (ie. just use `u16` everywhere). Users can still fetch the maximum sampling frequency manually and pass that.

PR #35, opened 2019-05-21
Benchmarks

djc:


spacejam:
Sorry for the delay, just got back from vacation. Looks good, thanks :D

PR #37, opened 2019-06-30
Add `--root` flag

### LINKED ISSUE #20: Flag for running perf or dtrace as root
It'd be handy to have a flag like `--root` that called `perf` and `dtrace` using `sudo` to avoid the need for changing system-wide permissions. It also (from memory) gives better visibility into kernel symbols, so a double win!

killercup:
This allows running this on macOS without having cargo write a bunch of
files as root.

Resolves #20

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -61,6 +61,10 @@ struct Opt {
     #[structopt(short = "f", long = "features")]
     features: Option<String>,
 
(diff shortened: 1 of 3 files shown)


alok:
@spacejam? I know you're probably busy but this would be very useful to me too after fighting permission errors trying to clean the target dir.

eliaslevy:
This would be a great addition.

eliaslevy:
Can we get it merged?

PR #42, opened 2019-09-19
Format code using 'cargo fmt'

Atul9:


jonas-schievink:
This was done as part of https://github.com/ferrous-systems/flamegraph/pull/58 and is now also tested in CI

PR #44, opened 2019-10-13
Use LBR instead of dwarf call-graph.

wathiede:
This option is viable on Haswell or later Intel platforms. It is much
lower overhead while profiling, generates more compact perf data, and is
much quicker to collapse and generate flame graphs.

The downsides is that it has limited stack depth, does not work on major
cloud providers and some others listed at
https://lwn.net/Articles/617097/ and
http://www.brendangregg.com/perf.html

Would you be interested in a change like this?  I'm also happy to make it configurable by env var of flag if you provide some guidance.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,7 +41,7 @@ mod arch {
     ) -> Command {
         let mut command = Command::new("perf");
 
-        for arg in "record -F 99 --call-graph dwarf -g"
+        for arg in "record -F 99 --call-graph lbr -g"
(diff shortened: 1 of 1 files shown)


spacejam:
Hey @wathiede, thank you for putting in the work to implement this. I think we should support allowing this, but I just implemented an option for users to supply an entire custom perf & dtrace invocations, for even more flexibility. In #86 I've also added an example to the README showing how to use LBR call graphs.

PR #45, opened 2019-10-26
Allow to pass in option --open to automatically open the resulting

VersBinarii:
.svg file in the default program. Its similar operation to the `cargo doc --open`


Cargo.toml
```diff
@@ -23,3 +23,4 @@ inferno = "0.4"
 signal-hook = "0.1.8"
 structopt = "0.2.14"
 cargo_metadata = "0.7.1"
+opener = "0.4.1"
```
bjorn3: nit: missing trailing newline

src/bin/cargo-flamegraph.rs
```diff
@@ -283,8 +287,20 @@ fn main() {
...
 
+    let should_open = opt.open;
+
     flamegraph::generate_flamegraph_by_running_command(
         workload,
-        flamegraph_filename,
+        flamegraph_filename.clone(),
     );
+
+    if should_open && flamegraph_filename.exists() {
```
bjorn3: `generate_flamegraph_by_running_command` exists the program when it couldnt generate the file, so it is unnecessary to check if it was created.

src/bin/cargo-flamegraph.rs
```diff
@@ -283,8 +287,20 @@ fn main() {
         .take()
         .unwrap_or("flamegraph.svg".into());
 
+    let should_open = opt.open;
```
bjorn3: I think it is cleaner to just use `opt.open` in the if condition below instead of assigning it to a variable.

src/bin/cargo-flamegraph.rs
```diff
@@ -61,6 +61,10 @@ struct Opt {
     #[structopt(short = "f", long = "features")]
     features: Option<String>,
 
+    /// Open the output .svg file with default program
+    #[structopt(short = "O", long = "open")]
```
jonas-schievink: Does `-O` have precedence in other tools? Cargo only allows `--open`, and I think we should do the same here unless there's a good reason to prefer `-O`.

VersBinarii:
Hello,
Sorry, it got a little messy with the rebasing. Hope all is fine now.

src/bin/cargo-flamegraph.rs
```diff
@@ -299,8 +303,18 @@ fn main() {
...
-        &flamegraph_filename,
+        &flamegraph_filename.clone(),
         opt.root,
         opt.frequency,
     );
+
+    if opt.open {
+        if let Err(e) = opener::open(&flamegraph_filename) {
+            eprintln!(
+                "Failed to open [{}]. Error: {:?}",
```
jonas-schievink:
```suggestion
                "Failed to open [{}]. Error: {}",
```

src/bin/flamegraph.rs
```diff
@@ -67,8 +71,18 @@ fn main() {
...
-        flamegraph_filename,
+        flamegraph_filename.clone(),
         opt.root,
         opt.frequency,
     );
+
+    if opt.open {
+        if let Err(e) = opener::open(&flamegraph_filename) {
+            eprintln!(
+                "Failed to open [{}]. Error: {:?}",
```
jonas-schievink:
```suggestion
                "Failed to open [{}]. Error: {}",
```

src/bin/cargo-flamegraph.rs
```diff
@@ -299,8 +303,18 @@ fn main() {
 
     flamegraph::generate_flamegraph_for_workload(
         Workload::Command(workload),
-        &flamegraph_filename,
+        &flamegraph_filename.clone(),
```
jonas-schievink: That function takes any `AsRef<Path>` implementor, so the previous code should have been fine

```suggestion
        &flamegraph_filename,
```

src/bin/flamegraph.rs
```diff
@@ -67,8 +71,18 @@ fn main() {
 
     flamegraph::generate_flamegraph_for_workload(
         workload,
-        flamegraph_filename,
+        flamegraph_filename.clone(),
```
jonas-schievink:
```suggestion
        &flamegraph_filename,
```

jonas-schievink:
Great, thanks!

Issue #49, opened 2019-11-08
windows install error


zydxhs:
OS: win10
Rust: 1.38 stable amd64
cargo install flamegraph, I get error like this:

   Compiling inferno v0.4.1
   Compiling cargo_metadata v0.7.4
   Compiling flamegraph v0.1.13
error[E0425]: cannot find value `exit_status` in this scope
   --> d:\Program\cargo\registry\src\mirrors.ustc.edu.cn-61ef6e0cd06fb9b8\flamegraph-0.1.13\src\lib.rs:134:6
    |
134 |     !exit_status.success()
    |      ^^^^^^^^^^^ not found in this scope

error: aborting due to previous error

For more information about this error, try `rustc --explain E0425`.
error: Could not compile `flamegraph`.
warning: build failed, waiting for other jobs to finish...
error: failed to compile `flamegraph v0.1.13`, intermediate artifacts can be found at `C:\Temp\cargo-installwtC1Id`

Caused by:
  build failed

ConnorSkees:
I am also facing this issue on Windows 10.
Toolchain: `stable-x86_64-pc-windows-msvc`
Rustc: `rustc 1.39.0 (4560ea788 2019-11-04)`
```
error[E0425]: cannot find value `exit_status` in this scope
--> C:\.cargo\registry\src\github.com-1ecc6299db9ec823\flamegraph-0.1.13\src\lib.rs:134:6
    |
134 |    !exit_status.success()
    |     ^^^^^^^^^^^ not found in this scope

error: aborting due to previous error

For more information about this error, try `rustc --explain E0425`.
error: failed to compile `flamegraph v0.1.13`
```
I guess something went wrong when refactoring [here](https://github.com/ferrous-systems/flamegraph/commit/49b1ad87cfaab97fe39e6ba5b3fb199241cd74c7#diff-b4aea3e418ccdb71239b96952d9cddb6R132)

zydxhs:
base v0.1.13, Just make the following changes and it will compile:

position: flamegraph-0.1.13\src\lib.rs:134:6
!exit_status.success() -> !status.success()

ConnorSkees:
@zydxhs I additionally had trouble building because of `signal_hook`, do you not run into this when building?

zydxhs:
@ConnorSkees Don not use the master version, there are many compilation errors.

jonas-schievink:
This is now fixed on master.

PR #50, opened 2019-11-08
Fix building on Windows

ConnorSkees:
Prior to this commit, it was not possible to build on Windows because of signal_hook. This commit changes the way signal_hook is used to only be imported when on Unix-like systems.

Additionally, during [`49b1ad`](
https://github.com/ferrous-systems/flamegraph/commit/49b1ad87cfaab97fe39e6ba5b3fb199241cd74c7#diff-b4aea3e418ccdb71239b96952d9cddb6R132) a variable was mistakenly left unchanged for the Windows variant of a function, which caused the build to fail.

This commit should resolve https://github.com/ferrous-systems/flamegraph/issues/49.

I've no idea what signal_hook is being used for here; please let me know if removing it from Windows would remove the functionality of the program as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,8 @@ path = "src/bin/flamegraph.rs"
 
 [dependencies]
 inferno = "0.4"
-signal-hook = "0.1.8"
 structopt = "0.2.14"
 cargo_metadata = "0.7.1"
+
+[target.'cfg(unix)'.dependencies]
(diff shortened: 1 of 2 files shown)


jonas-schievink:
Duplicate of https://github.com/ferrous-systems/flamegraph/pull/46, so closing in favor of that (which I can hopefully merge shortly)

PR #51, opened 2019-11-13
Fix mangling of whitespace-bearing arguments

Ralith:
Comes up when e.g. trying to filter a Criterion benchmark binary that uses human-friendly benchmark names.


Ralith:
Rebased.

matklad:
Thanks!

PR #53, opened 2019-11-30
Profile by pid

ASankaran:
Sometimes it is preferable to be able to profile a running process instead of needing to restart the process under the profiler. This PR adds a `Workload` enum which can contain a command or a pid. Additionally a flag for passing in a pid instead of a command is provided for the `flamegraph` binary.

Usage:
```
flamegraph -p 2147
```

These changes do not break the existing cli interface.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -3,6 +3,8 @@ use std::path::{Path, PathBuf};
 
 use structopt::StructOpt;
 
+use flamegraph::Workload;
+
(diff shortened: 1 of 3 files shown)


jonas-schievink:
Thanks, this looks like a reasonable addition!

I think it would be good if the CLI tool would raise an error if `-p` is used alongside the trailing command. Right now the code will ignore the command and use the PID.

ASankaran:
Sounds good. The CLI now checks that trailing_arguments is empty before generating the Workload for the pid.

PR #56, opened 2019-12-02
Format code and make CI more pedantic

jonas-schievink:
(opened from my fork to see what happens to the GH actions)


jonas-schievink:
Looks like GH workflows don't run when they're modified in a fork, kinda makes sense. https://github.com/ferrous-systems/flamegraph/pull/57 does run them as expected.

PR #84, opened 2020-04-15
Add some basic info about flamegraphs to README.md

nsvedberg:
I've never used a flamegraph before and was confused on what exactly the x and y axes represented. The README has a good overview of using a flamegraph for people who are already familiar with how to read them, but it didn't answer the basic questions I had. I added some information about what the axes, widths, and colors mean. Hopefully this will be easy to find for people who are skimming the docs.

I'm no expert so if this info is wrong or could be worded in a clearer way, please amend it before pulling. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,10 +111,23 @@ added together. Then an SVG is generated showing the
 call stacks that were measured, widened to the proportion
 of all stack samples that contained them.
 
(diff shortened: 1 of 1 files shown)


spacejam:
Thanks for improving this!

bors r+

PR #87, opened 2020-05-29
Update inferno version v0.9.8 with bugfix for C++ and Rust templates.

SoftwareApe:
Now templates containing parentheses aren't truncated anymore but properly displayed.

`std::function<int (int, int)>::operator` isn't truncated to `std::function<int ` anymore.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,9 +273,9 @@ dependencies = [
 
 [[package]]
 name = "inferno"
-version = "0.9.6"
+version = "0.9.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


SoftwareApe:
Bildschirmfoto von 2020-05-29 22-31-52

SoftwareApe:
v0.9.8 also adds support for curly braced lambdas in C++ and doesn't have the regression of v0.9.7 on anonymous namespaces, e.g. (anonymous namespace)::foo, doesn't get truncated anymore.
Curly braces and anonymous namespaces

spacejam:
looks good, thank you! I'll include a release with this in a few minutes

PR #88, opened 2020-05-31
Fix 'extraneous argument' error on macOS

mciantyre:
See "The second one" in #72. I noticed

```
dtrace: invalid probe specifier profile-997 /pid == $target/ { @[ustack(100)] = count(); }: extraneous argument 'my_arg' ($1 is not referenced)
failed to sample program
```

when using `flamegraph-rs` on macOS:

```
sudo cargo flamegraph --bin=my_binary -- my_arg
```

It looks like we're passing the command string, `path/to/my_binary my_arg` as multiple arguments, prefixed by flag `-c`. We should pass them as a *single* argument, prefixed by flag `-c`, as if they were quoted:

```
sudo dtrace -c './path/to/my_binary my_arg' ...
```

Changing `.args()` to `.arg()` resolves the "extraneous argument" error.

macOS (10.15.5)
flamegraph-rs 0.3.0
dtrace 1.15

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -129,7 +129,7 @@ mod arch {
         match workload {
             Workload::Command(c) => {
                 command.arg("-c");
-                command.args(&c);
+                command.arg(&c.join(" "));
             }
             Workload::Pid(p) => {
                 command.arg("-p");


vlthr:
This PR did the trick for me when passing single arguments, but the issue returned when I tried passing quoted multi-word arguments. The best solution that I could find was to manually escape whitespace within multi-word args before passing the cmd string to dtrace.

My use case was running the following criterion benchmark:
```
cargo flamegraph --root --bench core_bench -- --profile-time 10 'json -> Document'
```

Under this PR, it is passed to dtrace as: (note how `json -> Document` is unquoted)
```
"sudo" "dtrace" "-x" "ustackframes=100" "-n" "profile-997 /pid == $target/ { @[ustack(100)] = count(); }" "-o" "cargo-flamegraph.stacks" "-c" "/path/to/core_bench-fdfeb48cd9a646fb --profile-time 10 json -> Document"
```

I figured the best way to work around it would be to properly shell quote the multi-word arguments before passing them along, but apparently dtrace somehow [strips the quotes](https://8thlight.com/blog/colin-jones/2017/02/02/dtrace-gotchas-on-osx.html) before running the command. 

With shell quoting of the `-c` arg: (the command string looks correct, but dtrace does not respect the quotes)
```
"sudo" "dtrace" "-x" "ustackframes=100" "-n" "profile-997 /pid == $target/ { @[ustack(100)] = count(); }" "-o" "cargo-flamegraph.stacks" "-c" "/path/to/core_bench-fdfeb48cd9a646fb --profile-time 10 \'json -> Document\'"
```

What I ended up with was simply escaping spaces inside each of the arguments, which worked for all of my use-cases.

```rust
match workload {
    Workload::Command(mut c) => {
        // dtrace does not preserve quoting in the string passed via -c, so to pass along
        // multi-word arguments we need to escape any whitespace within each arg.
        for arg in c[1..].iter_mut() {
            *arg = arg.replace(" ", "\\ ")
        }
        command.arg("-c");
        command.arg(c.join(" "));
    }
    Workload::Pid(p) => {
        command.arg("-p");
        command.arg(p.to_string());
    }
}
``` 

Escaping spaces (with no shell quoting) gives the following command: (works!)
```
"sudo" "dtrace" "-x" "ustackframes=100" "-n" "profile-997 /pid == $target/ { @[ustack(100)] = count(); }" "-o" "cargo-flamegraph.stacks" "-c" "/path/to/core_bench-fdfeb48cd9a646fb --profile-time 10 json\\ ->\\ Document"
```

mciantyre:
Closing in favor of #100.

PR #89, opened 2020-05-31
Don't spend time optimizing build deps, such as proc macros

joshtriplett:
A build of flamegraph spends time optimizing various proc macros and
their dependencies, when they'll only run at build time and never at
runtime. Building them without optimizations substantially speeds up
the build, without affecting runtime performance at all.

Before:
    Finished release [optimized] target(s) in 2m 35s

After:
    Finished release [optimized] target(s) in 1m 31s


spacejam:
thank you! I'll cut a release with this in a few minutes

PR #90, opened 2020-06-10
Restore default sample rate of 997

yoshuawuyts:
https://github.com/flamegraph-rs/flamegraph/pull/33#diff-b4aea3e418ccdb71239b96952d9cddb6L107 modified the sample rate from 997 to 99, which often isn't enough resolution to capture short-lived events. This restores the sample rate to `997`, which should provide more granularity without significantly increasing overhead. Thanks!


spacejam:
looks good, thank you! I'll cut a release with this in a few minutes

Issue #63, opened 2019-12-26
Feature request: Specifying path to `perf` through environment


oxalica:
It seems the command `perf` (also, `dtrace`) is hard-coded and should be contained in `PATH`.
If `perf` is installed anywhere else, we need to modify `PATH` before running it, which is not a good way and may break other programs.

Related: https://github.com/NixOS/nixpkgs/pull/76313

jonas-schievink:
Perhaps we should respect `PERF` and `DTRACE` environment variables if they're set?

djc:
Looks like this has been implemented in #92.

PR #92, opened 2020-06-27
respect PERF and DTRACE env vars

spitfire05:
As suggested in #63


spacejam:
looks good, thank you! I'll include this in a release in a few minutes :)

PR #95, opened 2020-07-31
Detect '-o' argument in custom command.

vitorenesduarte:
Previously, if argument `-o` was set in the custom command, `flamegraph` would fail to compute perf's output since it would look for the default `perf.data` file.

Now, we detect if the user sets perf's output file with `-o` and build the flamegraph from that file instead.

With this change, the following command now seems to work:
```
flamegraph -c "record -F 997 --call-graph dwarf -g -o my_perf.data" date
Fri Jul 31 12:02:05 CEST 2020
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.075 MB my_perf.data (7 samples) ]
```

Without this change, this is the error we get:
```
flamegraph -c "record -F 997 --call-graph dwarf -g -o my_perf.data" date
Fri Jul 31 12:03:08 CEST 2020
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.075 MB my_perf.data (7 samples) ]
thread 'main' panicked at 'unable to generate a flamegraph from the collapsed stack data: Io(Custom { kind: InvalidData, error: "No stack counts found" })', /home/vitor.enes/.cargo/registry/src/github.com-1ecc6299db9ec823/flamegraph-0.3.1/src/lib.rs:273:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@ mod arch {
         sudo: bool,
         freq: Option<u32>,
         custom_cmd: Option<String>,
-    ) -> Command {
+    ) -> (Command, Option<String>) {
         let perf = env::var("PERF")
             .unwrap_or_else(|_| "perf".to_string());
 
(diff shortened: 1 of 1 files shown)


spacejam:
Thanks for this fix! I'll be merging this into a new release in the next few minutes if bors is happy.

bors r+

spacejam:

vitorenesduarte:
Awesome, thanks!

PR #98, opened 2020-09-17
find right path of binary files

VincentUCLA:
I'm not able to find the right path of binary automatically -- as i've moved the binary file from the `deps` folder to the upper level of `release` folder, the problem was solved.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -246,13 +246,12 @@ fn workload(opt: &Opt) -> Vec<String> {
 
     if opt.dev {
         binary_path.push("debug");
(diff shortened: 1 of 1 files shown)


arnej:
I can run examples without issues here with the current master.

When I try your PR and run `cargo flamegraph --dev --example readme`, I get the following error:

```
Workload failed: No such file or directory
failed to sample program
```

The example source is in `examples/readme.rs` and the built example is in `target/debug/examples`.

What is your directory structure after you run the following?

`cargo clean && cargo build`

spacejam:
Hi @VincentUCLA, specifically, are you trying to find the binary for a bench binary? Now that #102 is merged, does master work for you?

PR #100, opened 2020-09-29
Escape space when passing commands to arguments

djc:
Both of the fixes proposed in #88 don't work for me:

```diff
diff --git a/src/lib.rs b/src/lib.rs
index a83fd94..dde4adb 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -129,7 +129,7 @@ mod arch {
         match workload {
             Workload::Command(c) => {
                 command.arg("-c");
-                command.args(&c);
+                command.arg(&c.join(" "));
             }
             Workload::Pid(p) => {
                 command.arg("-p");
```

or 

```diff

diff --git a/src/lib.rs b/src/lib.rs
index dde4adb..4acd0b2 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,7 +127,12 @@ mod arch {
         command.arg("cargo-flamegraph.stacks");
 
         match workload {
-            Workload::Command(c) => {
+            Workload::Command(mut c) => {
+                // dtrace does not preserve quoting in the string passed via -c, so to pass along
+                // multi-word arguments we need to escape any whitespace within each arg.
+                for arg in c[1..].iter_mut() {
+                    *arg = arg.replace(" ", "\\ ")
+                }
                 command.arg("-c");
                 command.arg(&c.join(" "));
             }
```

So I came up with a new version that at least seems to work for me. @vlthr would you mind checking my work? @spacejam I noticed the work in #88 has been sitting there for a while, but this issue makes flamegraph quite a pain to use on macOS. Please consider reviewing and releasing this (or something like it).

(Also I'm happy to help out with maintaining this project if you prefer.)

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -280,7 +280,7 @@ fn workload(opt: &Opt) -> Vec<String> {
     } else if let Some(ref bench) = opt.bench {
         find_binary("bench", &binary_path, bench)
     } else if let Some(ref bin) =
-        opt.bin.as_ref().or(opt.example.as_ref())
+        opt.bin.as_ref().or_else(|| opt.example.as_ref())
     {
         if targets.contains(&bin) {
             bin.to_string()
@@ -324,7 +324,7 @@ fn main() {
     let flamegraph_filename: PathBuf = opt
         .output
         .take()
-        .unwrap_or("flamegraph.svg".into());
+        .unwrap_or_else(|| "flamegraph.svg".into());
 
     flamegraph::generate_flamegraph_for_workload(
         Workload::Command(workload),
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -71,6 +71,10 @@ struct Opt {
     #[structopt(long = "root")]
     root: bool,
 
+    /// Print extra output to help debug problems
+    #[structopt(short = "v", long = "verbose")]
+    verbose: bool,
+
     /// Sampling frequency
     #[structopt(short = "F", long = "freq")]
     frequency: Option<u32>,
@@ -127,6 +131,10 @@ fn build(opt: &Opt) {
     if let Some(ref features) = opt.features {
         cmd.arg("--features");
         cmd.arg(features);
+    }
+
+    if opt.verbose {
+        println!("build command: {:?}", cmd);
     }
 
     let mut child = cmd
@@ -309,6 +317,9 @@ fn main() {
     build(&opt);
 
     let workload = workload(&opt);
+    if opt.verbose {
+        println!("workload: {:?}", workload);
+    }
 
     let flamegraph_filename: PathBuf = opt
         .output
@@ -321,6 +332,7 @@ fn main() {
         opt.root,
         opt.frequency,
         opt.custom_cmd,
+        opt.verbose,
     );
 
     if opt.open {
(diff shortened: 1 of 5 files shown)


vlthr:
This works great for me! (macOS 10.15.4 Catalina)

alexhexabeam:
This works for me as well, on macOS 10.14.6 Mojave.

trevyn:
Also confirming that this fixes the arg-passing issue for me on 10.14.6.

May it save you a few seconds until this is merged:

`cargo install --git "https://github.com/djc/flamegraph" --branch dtrace-args`

src/lib.rs
```diff
@@ -137,8 +132,17 @@ mod arch {
...
+                    if i > 0 {
+                        escaped.push(' ');
+                    }
+                    escaped
+                        .push_str(&arg.replace(" ", "\\ "));
+                }
+
                 command.arg("-c");
-                command.args(&c);
+                command.arg(&escaped);
```
michaelkirk, line 145:
```suggestion
                command.arg(c.join(" "));
```

I think we can omit all of the above and pass all of c as a single arg.

src/bin/flamegraph.rs
```diff
@@ -75,14 +79,15 @@ fn main() {
     let flamegraph_filename: PathBuf = opt
         .output
         .take()
-        .unwrap_or("flamegraph.svg".into());
+        .unwrap_or_else(|| "flamegraph.svg".into());
```
michaelkirk, line 82: why this change?

src/bin/cargo-flamegraph.rs
```diff
@@ -309,18 +317,22 @@ fn main() {
     build(&opt);
 
     let workload = workload(&opt);
+    if opt.verbose {
```
michaelkirk, line 320: At least one of the dependencies `inferno`, already relies on `log`, maybe better to use that instead of introducing our own log level machinery.

michaelkirk (commented):
(Just a drive by reviewer - I have no authority here.)

This fix worked for me as is, though I've suggested a change for a smaller diff.

And not sure about the unrelated changes, but I'll leave that to the maintainer(s). 

Thanks for pushing up this fix @djc!

src/bin/flamegraph.rs
djc: Note that this is in a separate commit, with the commit message noting it cleans up some clippy warnings. (clippy warns against the before pattern because it performs a potentially expensive operation regardless of whether it's needed or not.)

kdy1:
I merged this and #102 in my fork.

See: https://github.com/kdy1/flamegraph

kevinkassimo:
Bump this PR again for maintainer attention. This problem is quite annoying.

djc:
(I sent the maintainer an email last week to ask if I could help out with maintaining this crate.)

kdy1:
I created an alternative tool which can do `cargo flamegraph` can do.
See: https://github.com/kdy1/cargo-profile
It also allows profiling cpu usage per function on mac os x.

spacejam:
thanks for your patience folks, I've returned after taking a few months away from the computer to read and spend time with humans :) I'll be cutting a release shortly with the latest batch of updates

bors r+

spacejam:
Version 0.4 has been released which includes these changes!

And welcome @djc, the newest maintainer, to the team :)

djc:
Thanks!

Issue #32, opened 2019-04-25
unable to collapse generated profile data: Custom { kind: InvalidData, error: StringError("stream did not contain valid UTF-8") }


jjyr:
Randomly panic on my mac

```
thread 'main' panicked at 'unable to collapse generated profile data: Custom { kind: InvalidData, error: StringError("stream did not contain valid UTF-8") }', src/libcore/result.rs:997:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::_print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::result::unwrap_failed
   9: flamegraph::generate_flamegraph_by_running_command
  10: cargo_flamegraph::main
  11: std::rt::lang_start::{{closure}}
  12: std::panicking::try::do_call
  13: __rust_maybe_catch_panic
  14: std::rt::lang_start_internal
  15: main
```

lmcgrath:
Happening on mine too for all runs
```
thread 'main' panicked at 'unable to collapse generated profile data: Custom { kind: InvalidData, error: "stream did not contain valid UTF-8" }', /Users/logan/.cargo/registry/src/github.com-1ecc6299db9ec823/flamegraph-0.3.0/src/lib.rs:236:5
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::panicking::default_hook::{{closure}}
   4: std::panicking::default_hook
   5: std::panicking::rust_panic_with_hook
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::option::expect_none_failed
   9: flamegraph::generate_flamegraph_for_workload
  10: cargo_flamegraph::main
  11: std::rt::lang_start::{{closure}}
  12: std::rt::lang_start_internal
  13: main
```

michaelkirk:
I hit this too. I'm not sure yet what's causing it, but here are some notes:

First, I commented out the line that deletes the raw dtrace dump:

```patch
@@ -163,11 +166,11 @@ mod arch {
              output file cargo-flamegraph.stacks",
         );
 
[...]

PR #101, opened 2020-10-02
Workaround #32 - fails parsing invalid utf8 dtrace output (macos only?)

michaelkirk:
Intermittently, invalid utf-8 is found in cargo-flamegraph.stacks, which causes parsing to blow up with the error:

> unable to collapse generated profile data: Custom { kind: InvalidData, error: StringError("stream did not contain valid UTF-8") }

This commit doesn't fix the underlying problem, but simply works around
it by lossily re-encoding to valid utf8.

Anecdotally it seems to be macos symbol names at the end of the line that
contain the invalid utf-8 - I don't know if this is due to some error in dtrace
or if somehow the symbols actually contain non utf-8 encodings.

Note I did try explicitly specifying utf8 output, by adding to the
dtrace command invocation:

    command.arg("-x");
    command.arg("encoding=utf8");

But I ran into the same error seemingly just as often.

---

This is admittedly a hack, so I understand if you don't want to merge it, but it might be helpful for folks like me experiencing #32. 

Anecdotally this commit seems to completely fix things for me. Without it I get the above error about 50% of the time — making it quite frustrating to use this otherwise very nice tool. 🙂 

The caveat is that presumably the invalid symbol names will not be correctly labeled/classified, but in practice this hasn't bitten me yet, since it seems to be a relatively small number of affected lines.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -169,7 +169,22 @@ mod arch {
                  temporary file",
             );
 
-        buf
+        // Workaround #32 - fails parsing invalid utf8 dtrace output
+        //
+        // Intermittently, invalid utf-8 is found in cargo-flamegraph.stacks, which
+        // causes parsing to blow up with the error:
+        //
+        // > unable to collapse generated profile data: Custom { kind: InvalidData, error: StringError("stream did not contain valid UTF-8") }
+        //
+        // So here we just lossily re-encode to hopefully work around the underlying problem
+        let string = String::from_utf8_lossy(&buf);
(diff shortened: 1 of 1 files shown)


spacejam:
Thanks for the workaround! I have also been bitten by this on OSX and I think this is an acceptable workaround for the time being. I've been bitten by a large number of annoying filesystem issues on OSX over time...

I'll be cutting a new release in a few minutes that includes this change if bors is happy.

bors r+

spacejam:

PR #102, opened 2020-10-05
Append 'deps' to binary_path when running benches

arnej:
I was only able to test this for libraries (I tested [comrak](https://crates.io/crates/comrak)).

Without this, no benches are found with `cargo +nightly flamegraph --bench progit`.

With this change, the benchmark runs fine and a flamegraph got created.


JAicewizard:
this also works properly for me (linux).
would be super amazing to have this merged!

spacejam:
thanks for the fix! I'll be cutting a release with this shortly

bors r+

spacejam:

PR #103, opened 2020-10-20
Read target triple from environment

pikajude:
When `$CARGO_BUILD_TARGET` or `--target` is specified, the build artifacts end up in the directory `./target/$target_triple`. Use case: generating flamegraphs for `x86_64-unknown-linux-musl` on a GNU linux build machine.


pikajude:
bors try

spacejam:
Thanks for this update!

I'll cut a new release that includes this change in a few minutes if bors is happy.

bors r+

spacejam:

PR #107, opened 2020-12-07
Cut 0.4

spacejam:


polyzen:
Please start flagging tags again so downstream packagers can be notified when following the repo or [Atom feed](https://github.com/flamegraph-rs/flamegraph/releases). Including release notes as you had been doing is also a treat. ^^

djc:
I've pushed tags for 0.3.1 and 0.4.0, and also published some brief changelogs.

polyzen:
Thank you very much.

Issue #114, opened 2021-01-06
Support cargo `--no-default-features`


ashpil:
Currently, flamegraph supports the `--features` cargo flag, however, I don't think there's any way to work with that and disable default features. It would be nice if you could disable default features when running flamegraph (and this would be extraordinarily simple to implement). 

It might be even better to have some sort of cargo passthrough, where one can specify a `--cargo-cmd="--no-default-features"` or something that would pass any given command to cargo.

djc:
If you implement it, I'll review it. I think I'd prefer to stick to just `--no-default-features` for now, rather than a passthrough option.

ashpil:
See #115.

PR #115, opened 2021-01-06
Add support for --no-default-features

ashpil:
This implements #114.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -67,6 +67,10 @@ struct Opt {
     #[structopt(short = "f", long = "features")]
     features: Option<String>,
 
+    /// Disable default features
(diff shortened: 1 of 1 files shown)


ashpil:
@djc

src/bin/cargo-flamegraph.rs
```diff
@@ -67,6 +67,10 @@ struct Opt {
     #[structopt(short = "f", long = "features")]
     features: Option<String>,
 
+    /// Disable default features
+    #[structopt(short = "n", long = "no-default-features")]
+    no_def_features: bool,
```
djc, line 72: Let's not have a short option for this (Cargo doesn't either), and let's expand the field to the full name.

ashpil: Done

djc:
Thanks!

PR #116, opened 2021-01-14
Recommend CARGO_PROFILE_RELEASE_DEBUG

chinedufn:
RUSTFLAGS does not always work https://github.com/rust-lang/rust/issues/71848#issuecomment-623159587


djc:
bors r+

Thanks!

PR #118, opened 2021-01-22
Add --image-width

### LINKED ISSUE #109: Function names truncated
The function signatures I'm working with are rather long and get truncated in the resulting flamegraph to the point of not being helpful. Is there anyway to make the outputted svg wider to combat this? Thanks!

0:
Fixes #109 by adding the command-line option `--image-width`. The specified value is passed directly to Inferno's [`Options`](https://docs.rs/inferno/0.10.2/inferno/flamegraph/struct.Options.html#structfield.image_width).


djc:
LGTM, thanks!

bors r+

PR #120, opened 2021-02-11
Incorporate more of inferno-flamegraph's CLI

saethlin:


src/lib.rs
```diff
@@ -322,3 +316,59 @@ pub fn generate_flamegraph_for_workload<
...
+    #[structopt(long = "notes", value_name = "STRING")]
+    pub notes: Option<String>,
+
+    /// Omit functions smaller than <FLOAT> pixels
+    #[structopt(
+            long = "minwidth",
+            default_value = &defaults::str::MIN_WIDTH,
+            value_name = "FLOAT"
+        )]
+    pub minwidth: f64,
```
djc, line 350: Let's call this `min_width` in the code and `min-width` in the CLI?

djc (commented):
LGTM, one naming nit:

src/lib.rs
saethlin: My initial stab at this was to paste the options from the `inferno` series of pipes onto this crate.

However, if we're fixing this spelling inconsistency (that is, an `inferno` user won't be able to just use the flags they're used to if we do this), do we want to also adjust the defaults? I don't know what other users experiences are like, but I've never been in a situation where I want nondeterministic colors when comparing data visualizations (it's my understanding that `inferno` inherits this as legacy behavior from the Perl program), and I've never found the file size vs information tradeoff at `min_width = 0.1` to be the right one; a factor of 10 less preserves a lot of useful context. I suspect this 0.1 value was chosen based on work on software that doesn't spawn tens of threads.

djc: For this PR, let's just change the spelling a little bit. I'm open to changing the defaults if we think they can be more useful, but I'd prefer to do that in separate PRs (in which case, please include screenshots).

djc (approved):

djc:
Bors r+

djc:
Thanks!

PR #121, opened 2021-02-16
support rustc --manifest-path option

eugene-babichenko:
This is makes running `cargo flamegraph` for benchmarks from the workspace root very similar to `cargo bench`.


djc:
Thanks! Bors r+

eugene-babichenko:
Is bors ignoring us? :(

djc:
I guess it needs to be on a line by itself?

@bors r+

djc:
Uhh...

bors r+

PR #122, opened 2021-02-16
use the `bench` profile for building benchmarks

eugene-babichenko:
This will remove the possibility to build benchmarks with the `dev`
profile, since `cargo bench` does not support `dev`.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -117,9 +117,17 @@ enum Opts {
 
 fn build(opt: &Opt) {
     let mut cmd = std::process::Command::new("cargo");
-    cmd.arg("build");
 
-    if !opt.dev {
+    // This will build benchmarks with the `bench` profile. This is needed
+    // because the `--profile` argument for `cargo build` is unstable.
+    if !opt.dev && opt.bench.is_some() {
+        cmd.args(&["bench", "--no-run"]);
+    } else {
+        cmd.arg("build");
+    }
+
+    // do not use `--release` when we are building for `bench`
+    if !opt.dev && opt.bench.is_none() {
         cmd.arg("--release");
     }
 


src/bin/cargo-flamegraph.rs
```diff
@@ -113,10 +114,18 @@ enum Opts {
...
     let mut cmd = std::process::Command::new("cargo");
-    cmd.arg("build");
 
-    if !opt.dev {
-        cmd.arg("--release");
+    if let Some(ref bench) = opt.bench {
+        cmd.arg("bench");
+        cmd.arg("--no-run");
+        cmd.arg("--bench");
+        cmd.arg(bench);
```
djc, line 122: I find `bench --no-run` pretty unintuitive. It seems like we could do `build --profile bench` instead?

djc:
(Also, this needs to be rebased.)

src/bin/cargo-flamegraph.rs
eugene-babichenko: Could you please check it on your setup? I have the following error.

```
❯ cargo build --profile bench
error: Usage of `--profile` requires `-Z unstable-options`
```

eugene-babichenko:
```
❯ cargo --version
cargo 1.50.0 (f04e7fab7 2021-02-04)
```

djc: Yeah, you're right -- weird that it shows it in the `--help` even when it's not actually available.

djc: Okay, in that case can you just add a comment here to make it clearer what this is trying to do (namely, building with the `bench` profile)?

eugene-babichenko: Yes, sure.

eugene-babichenko:
I am also thinking whether it would be practical to allow building benchmarks with `dev` profiles using the old `cargo build` approach? Also will remove breakage from the command interface (`--bench` is incompatible with `--dev` with this change).

djc:
I'm not really clear on what you mean/how you think this could be achieved. Can you elaborate a bit?

eugene-babichenko:
> what you mean

`cargo bench` can only build with the `bench` profile. Using `release` or `dev` is impossible using this command. Because of that I marked the `--dev` flag of `cargo-flamegraph` as incompatible with `--bench`. Basically I'm just asking if we want to return the possibility to have `cargo flamegraph --dev --bench` or just leave it as it is in this PR.

> how you think this could be achieved

By falling back to `cargo build --bench` when we have both `--dev --bench`.

djc:
Let me check my understanding: your main goal with this PR is to have `--bench` invocations compile with the bench profile rather than the release profile, but currently the only way to do that is to run the `cargo bench` command directly.

If we can avoid failing on `--dev --bench` (and if we can execute something meaningful for these options that matches the likely intent of the user), then we should probably keep that combination working.

Does that make sense?

eugene-babichenko:
Yes, absolutely. Then I will take my time to keep `--dev --bench` working.

src/bin/cargo-flamegraph.rs
```diff
@@ -117,7 +117,20 @@ enum Opts {
...
+        // This will build benchmarks with the `bench` profile. This is needed
+        // because the `--profile` argument for `cargo build` is unstable.
+        cmd.arg("bench");
+        cmd.arg("--no-run");
+    }
+
+    if let Some(ref bench) = opt.bench {
+        cmd.arg("--bench");
+        cmd.arg(bench);
+    }
```
djc, line 133: Would you mind squashing the commits and leaving the `if let Some(ref bench) = opt.bench {}` is its former place? That would make for a nicely minimal change which I think solves your issue.

eugene-babichenko: Did that. Slightly rewrote the whole thing actually.

djc:
LGTM, thanks for sticking with it!

bors r+

PR #124, opened 2021-02-19
fix false alarm about not having debuginfo in benchmarks

### LINKED ISSUE #123: False `building without debuginfo` warning when running benchmarks
When running benchmarks there is always a false `building without debuginfo` even when they are certainly enabled and reflected in a flamegraph. This must not happen since this is very misleading.

This is most likely due to [this condition](https://github.com/flamegraph-rs/flamegraph/blob/master/src/bin/cargo-flamegraph.rs#L199-L202) that expects all binaries names to end with a target name. In practice, benchmark binaries names are in format `bench-deadbeef` like intermediate build artifacts.

eugene-babichenko:
Fixes #123

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -195,6 +195,17 @@ fn build(opt: &Opt) {
 
         let mut has_debuginfo = false;
 
+        // get names of binaries in the workload
+        let workload_filenames = workload(opt)
+            .iter()
+            .filter_map(|w| {
+                PathBuf::from(w)
+                    .file_name()?
+                    .to_str()
+                    .map(|filename| filename.to_owned())
+            })
+            .collect::<Vec<_>>();
+
         // This is an extremely coarse check to see
         // if any of our build artifacts have debuginfo
         // enabled.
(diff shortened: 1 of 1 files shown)


src/bin/cargo-flamegraph.rs
```diff
@@ -210,25 +230,21 @@ fn build(opt: &Opt) {
...
-            if workload(opt)
-                .iter()
-                .any(|w| w.ends_with(&artifact.target.name))
-                && artifact.profile.debuginfo.unwrap_or(0)
-                    != 0
+            // - Check that this is a binary we are interested in.
+            //   This should start with the target name (binaries can have names in format
+            //   `benchmark-deadbeef`)
+            // - The iterator is reversed because the entities we are interested in are most likely
+            //   to appear in the of the list.
```
djc, line 237: This sentence is missing something "are most likely to appear in the [?] of the list".

src/bin/cargo-flamegraph.rs
```diff
@@ -195,6 +195,26 @@ fn build(opt: &Opt) {
 
         let mut has_debuginfo = false;
 
+        let profile = if opt.bench.is_some() {
+            "bench"
+        } else {
+            "release"
+        };
```
djc, line 202: Moving this code seems unrelated? Would you mind putting it in a separate commit (explaining why this good)?

src/bin/cargo-flamegraph.rs
eugene-babichenko: In the beginning of the list. Thanks for noting that.

eugene-babichenko:
Fixed everything.

src/bin/cargo-flamegraph.rs
```diff
@@ -195,6 +195,20 @@ fn build(opt: &Opt) {
...
+        let workload = workload(opt)
+            .iter()
+            .map(|w| {
+                w.parse::<PathBuf>()
+                    .unwrap()
+                    .file_name()
+                    .unwrap()
+                    .to_str()
+                    .unwrap()
+                    .to_owned()
```
djc, line 214: So I'm still a little bit stuck on this thing... Can we do something less finicky here? It's pretty confusing what's going on here, with a lot of unwrapping. Could we make this `filter_map()` and use lots of `ok()?` to just ignore the failures? I think we could do `PathBuf::from(w)` instead of `w.parse::<PathBuf>().unwrap()`?

eugene-babichenko: >  think we could do `PathBuf::from(w)`

Yes, good point.

> Could we make this `filter_map()` and use lots of `ok()?`

Yeah, makes sense, would more reliable. I am not exactly sure if this could be made simpler, we are going to use all of these operations anyway to get the file name.

djc (approved):
Thanks!

Issue #126, opened 2021-02-23
Add usage instructions


fabianbergmark:
It's pretty unclear how to actually use this tool.

I've guessed this much:

cargo flamegraph --bin BINARY

press ctrl-c after a while, I see

[ perf record: Woken up 1096 times to write data ]
[ perf record: Captured and wrote 275.761 MB perf.data (34147 samples) ]

now it just hangs, until I press ctrl-c again 
then I see

writing flamegraph to "flamegraph.svg"

It also seems that depending on how fast i issue the second ctrl-c, I get more or less output...

RandomInsano:
That's largely how I use it on macOS and it's great. Linux I have yet to get working properly.

djc:
If there is anything unclear about the usage instructions in the README, PRs are welcome. If you think there are bugs, then maybe make a more specific issue about the bug you think you're seeing.

fabianbergmark:
Provided that I'm not sure that I've ever been able to run this tool correctly, I'm not very inclined to write a PR for it.

djc:
That is totally understandable. (Although #127 suggests it might fix the issues you're having.)

RandomInsano:
I gave it a shot, but I'm in a different boat. No dice on my side. I'm still just left with perf.data

```
$ cargo flamegraph --root --no-inline --bin game -o flamegraph.svg
    Finished release [optimized] target(s) in 0.21s

WARNING: building without debuginfo. Enable symbol information by adding the following lines to Cargo.toml:

[profile.release]
debug = true

Error: "Exiting"
[ perf record: Woken up 36 times to write data ]
[ perf record: Captured and wrote 9.889 MB perf.data (1217 samples) ]
failed to sample program
$ ls -l
total 10152
-rw-rw-r-- 1 me me      786 Feb 20 21:01 Cargo.toml
drwxrwxr-x 2 me me     4096 Feb 14 23:13 examples
-rw------- 1 root        root        10377460 Mar  2 19:37 perf.data
drwxrwxr-x 6 me me     4096 Feb 14 23:13 Resources
drwxrwxr-x 7 me me     4096 Feb 14 23:13 src
```

I don't think this is a documentation issue because I've been using cargo-flamegraph for at least nine months on Mac with no trouble.

djc:
If you can run with `-v`, it should show the exact perf command it is running. Maybe try to debug that separately?

Aranjedeath:
[...]

PR #127, opened 2021-02-24
Add no-inline option

### LINKED ISSUE #74: addr2line taking an exorbitant amount of time
Hi,
I've recently been using flamegraph on a Linux system and once the recording is done it takes an extremely long time. It seems to be invoking addr2line over and over again making the whole process quite slow.

The output is:
```
[ perf record: Woken up 682 times to write data ]
[ perf record: Captured and wrote 170,633 MB perf.data (21198 samples) ]
```

21k samples don't sound that much but if it's invoking a program for every sample it that seems to become very expensive.

inteon:
Fixes #74 and #126

This is how to use it:
```sh
flamegraph --no-inline -o flamegraph.svg [COMMAND]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ usually something like `~/.cargo/bin`.
 # if you'd like to profile an arbitrary executable:
 flamegraph [-o my_flamegraph.svg] /path/to/my/binary --my-arg 5
 
+# NOTE: By default, perf tries to compute which functions are
+# inlined at every stack frame for every sample. This can take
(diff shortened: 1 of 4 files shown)


djc:
Would it be useful to add some hints to the README about how/when to use this?

Additionally I'd like to error out when this option is used on macOS (that is, when the backend doesn't support the option).

Licenser:
<3 this would be lovely to have!

inteon:
@djc I added the requested changes. As I don't use macOS, I was not able to test that yet.

inteon:
@djc @Licenser Can this be merged?

PR #134, opened 2021-05-01
Add --lang and --colors cli parameters

### LINKED ISSUE #19: Allow user to specify language for coloring flame graph
Inferno supports setting a color palette for the flame graph (just like `flamegraph.pl --colors`), and some of the palettes (see [`MultiPalette`](https://docs.rs/inferno/0.5.0/inferno/flamegraph/color/enum.MultiPalette.html)) are specifically designed to highlight particular languages well. It'd be cool if `cargo flamegraph` let you pass a `--lang` parameter to select the language to color the flame graph as! Speaking of, it'd be _really_ neat if we also had support for a Rust coloring that, say, colors functions from `std` and `core` differently from user functions!

zthompson47:
Fixes #19

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,7 @@ use inferno::collapse::dtrace::{
 
 use inferno::{
     collapse::Collapse,
+    flamegraph::color::Palette,
     flamegraph::{defaults, from_reader},
 };
 
@@ -367,6 +368,27 @@ pub struct FlamegraphOptions {
     /// Image width in pixels
     #[structopt(long = "image-width")]
     pub image_width: Option<usize>,
+
+    /// Color palette
+    #[structopt(
+        long = "colors",
+        short = "C",
+        conflicts_with = "lang",
+        possible_values = &[
+            "hot", "mem", "io", "red", "green", "blue", "aqua", "yellow",
+            "purple", "orange", "wakeup"
+        ]
+    )]
+    pub colors: Option<Palette>,
+
+    /// Color palette based on language semantics
+    #[structopt(
+        long = "lang",
+        short = "l",
+        conflicts_with = "colors",
+        possible_values = &["java", "perl", "js"]
+    )]
+    pub lang: Option<Palette>,
 }
 
 impl FlamegraphOptions {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -367,6 +368,16 @@ pub struct FlamegraphOptions {
     /// Image width in pixels
     #[structopt(long = "image-width")]
     pub image_width: Option<usize>,
+
+    /// Color palette
+    #[structopt(
+        long = "palette",
+        possible_values = &[
```
djc, line 375: What's the value of listing all possible values here? A downside is that we have to keep this list in sync with inferno.

djc (commented):
Mostly looks good, one question remaining for me. Thanks for working on this!

src/lib.rs
zthompson47: That's a good question, and TBH I listed them simply because they're available, thinking it would be nice to have as many options as possible for the `flamegraph` and `cargo-flamegraph` binaries.  There are some general reasons for having multiple palettes available, including:
* choice in case some palettes are more or less readable for a particular end-user
* profiling an app under different parameters and using separately colored flamegraphs for comparison

 I also thought that there was a way to automatically generate the list of palettes (using `clap::arg_enum!`), but I think that would have to be implemented in the inferno repo.

I guess I don't have the full picture of how this tool is used, but it appears to be the main cli frontend to the inferno lib, and would therefore want to harness the full capabilities of the backend.

As far as maintainability, I suppose we would still need to manually update the `possible_values` any time the palettes change in inferno, whether or not we use the complete list.  For example, I have a pending PR for adding `--palette rust` that we're adding here, and I could see other variants added down the line.  Somehow I don't imagine variants being removed, and I suspect new variants would be rare.

Hopefully that all makes sense, and I'm looking forward to hearing if you have suggestions on how to proceed.  Thanks!

djc: I think you don't have to make the possible values explicit because the FromStr impl for Palette will take care of validation.

zthompson47: Yes the validation is indeed handled by the FromStr impl, and I tested it out over here just to make sure.  A better answer to your question is that using `possible_values` will display a list of options to the user when running `cargo flamegraph -h`.  So the value is in documentation and discovery.

djc:
bors r+

Thanks again!

Issue #139, opened 2021-08-04
--test fails with `error: no test target named`


ede1998:
I'm trying to run `cargo flamegraph` for the unit tests in my crate. Unfortunately, it always fails with this error:
`error: no test target named {value}`
I tried numerous variations of the `--test` flag in combination with the binary name, path, test name, test full names, ... but nothing seems to work.

Is the `--test` feature not working right now or am I using it wrong? If I am using it wrong, can we maybe get a usage example in the README?

Workaround:
I manually build the test executable with `cargo build --tests --release` and then ran `flamegraph "target/release/deps/test-executable-9fb133aef8031813 -- full::test::name`

djc:
I think it does work to some extent, or there would likely be more issues filed. Are you using the latest version? Would you mind digging in a little to figure out why it's failing? You could try to pass `-v` to get verbose output that prints the build command.

ede1998:
I think I found the reason:
`cargo flamegraph --test <name>` invokes `cargo build --release --test <name>`, which fails if you invoke it with a unit test name. [The Cargo Book](https://doc.rust-lang.org/cargo/commands/cargo-build.html) says that `--test <name>` is used to "build the specified integration test". It seems to me like `cargo flamegraph` cannot run unit tests right now and the workaround I described above is needed.
After looking a bit into the `cargo build` documentation, I didn't find a way to use `cargo build` to just build the unit test executable. Is this even possible? `--tests` builds all tests, integration tests and unit tests alike.

Answering your questions:

> Are you using the latest version?

I think so, just installed it yesterday using cargo. `cargo flamegraph --version` prints `0.4.0`.

> Would you mind digging in a little to figure out why it's failing?

I created a test project to reproduce it: `cargo new flamegraph-test`. I added a unit test to `main.rs`:
```rust
use std::time::Duration;

fn main() {
    println!("Hello, world!");
}

#[test]
fn test_run() {
    std::thread::sleep(Duration::from_secs(3));
}
```
When I run `cargo flamegraph --test test_run`, I get the error `error: no test target named 'test_run'`.

> You could try to pass -v to get verbose output that prints the build command.

After adding the `-v` flag, flamegraph prints `build command: "cargo" "build" "--release" "--test" "test_run"`. Trying to run this command directly yields the same `error: no test target named 'test_run'` error.

djc:
[...]

PR #140, opened 2021-08-11
Profile unit tests

ede1998:
## About
This PR adds a feature to `cargo flamegraph`. It adds a flag `--unit-test` to the available command line options which profiles a unit test executable. The previous version only allowed profiling of integration tests. The flag is mutually exclusive to all other profiling target selection options (i.e. `--bin`, `--bench`, `--test`, `--example`). It takes an optional value that specifies the target that should be unit-tested. The target is required if the package has multiple targets.

## Relevant issue:
- #139 

### Unit test selection often requires -- as separator
Whenever the new flag is passed as the last flag, it is necessary to separate the following test selection like so:
`cargo flamegraph --unit-test -- test::name`
The reason behind this is that `--unit-test` takes an optional argument. `clap` tries to consume the test selection arguments as values for `--unit-test` instead. This leads to either a parse error or an error when trying to find the test selection as target.

This could be prevented by removing the optional argument to `--unit-test` and adding another flag `--unit-test-target <target_name>` but I'm not convinced that this improves the overall situation. Another solution might be to redesign the command line interface entirely and split it into subcommands:
```
cargo flamegraph bin
cargo flamegraph --dev test integration_test
cargo flamegraph --no-inline --package package1 unittest --target target_name test::name
```
This, however, seems to be too big of a change as it would be quite drastic and a breaking change.

### `test --no-run` should be used instead of `build --tests` (Obsolete, fixed in commit 80e47f4b64f5bf189d62e002edd8091f1826f04c)
Currently, the code invokes `cargo build --tests` to build the unit test executable before profiling. This leads to two issues:
1. It builds most tests (by default all unit and integration tests) leading to increased compile times and unrelated compiler errors blocking the profiling.
2. If the user has `test = false` for a target, it **does not** build the test executable, which means we cannot profile it.

Both issues can be resolved by using `cargo test --no-run --lib` or `cargo test --no-run --bin bname` to compile the test executable. But it requires knowing whether we are testing a lib or a binary and in the binary case the name is needed as well.
Therefore I decided not to implement it yet. I plan to tackle this issue after a second PR that changes `cargo flamegraph` to parse JSON output of the build commands instead of searching for the binary by name (created by now: #141).

### Wrong target binary selection (Obsolete, fixed with merge of  #141)
Using the current approach of searching for the binary by name, it is impossible to tell if we have found the unit test binary or the actual binary:
```
$ cargo build --release; cargo build --release --tests
$ ls target/release/deps
main1-7d658e30325341e3 main1-65b88a055fb7b8c6
```
Simply taking the binary with the newest timestamp can alleviate the situation in most cases. But to make it foolproof, it's necessary to read the output of cargo build. Therefore choosing by timestamp is not implemented as well. The aforementioned second PR will deal with this issue.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -101,7 +101,7 @@ enum Opts {
     Flamegraph(Opt),
 }
 
-fn build(opt: &Opt) -> anyhow::Result<Vec<Artifact>> {
+fn build(opt: &Opt, kind: impl IntoIterator<Item = String>) -> anyhow::Result<Vec<Artifact>> {
     use std::process::{Command, Output, Stdio};
     let mut cmd = Command::new("cargo");
 
@@ -109,6 +109,8 @@ fn build(opt: &Opt) -> anyhow::Result<Vec<Artifact>> {
     // because the `--profile` argument for `cargo build` is unstable.
     if !opt.dev && opt.bench.is_some() {
         cmd.args(&["bench", "--no-run"]);
+    } else if opt.unit_test.is_some() {
+        cmd.args(&["test", "--no-run"]);
     } else {
         cmd.arg("build");
     }
@@ -143,8 +145,11 @@ fn build(opt: &Opt) -> anyhow::Result<Vec<Artifact>> {
         cmd.arg(bench);
     }
 
-    if opt.unit_test.is_some() {
-        cmd.arg("--tests");
+    if let Some(Some(ref unit_test)) = opt.unit_test {
+        match kind.into_iter().any(|k| k == "lib") {
+            true => cmd.arg("--lib"),
+            false => cmd.args(&["--bin", unit_test]),
+        };
     }
 
     if let Some(ref manifest_path) = opt.manifest_path {
@@ -254,6 +259,7 @@ fn workload(opt: &Opt, artifacts: &[Artifact]) -> anyhow::Result<Vec<String>> {
 struct BinaryTarget {
     package: String,
     target: String,
+    kind: Vec<String>,
 }
 
 impl std::fmt::Display for BinaryTarget {
@@ -278,6 +284,7 @@ fn find_unique_target(kind: &[&str]) -> anyhow::Result<BinaryTarget> {
                     .then(|| BinaryTarget {
                         package: name.clone(),
                         target: t.name,
+                        kind: t.kind,
                     })
             })
         })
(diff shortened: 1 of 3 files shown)


ede1998:
@djc  I'll go over this one again this afternoon probably and clean up the commit history as well as incorporate the changes from #142 that are fixing the known issues of this PR.

djc:
Sounds good! I just pushed some more tweaks to your code and changed the approach to error handling. I hope this doesn't cause too much extra work in rebasing this (hopefully it would make things easier), but please rebase on current main.

ede1998:
Ok, I ended up mostly rewriting the PR but I could reuse some snippets. I tried to have a cleaner commit history this time.
80e47f4b64f5bf189d62e002edd8091f1826f04c may be a bit big though and it changes much of the previous work in this PR.

src/bin/cargo-flamegraph.rs
```diff
@@ -100,51 +100,104 @@ enum Opts {
     Flamegraph(Opt),
 }
 
-fn build(opt: &Opt) -> anyhow::Result<Vec<Artifact>> {
-    use std::process::{Command, Output, Stdio};
-    let mut cmd = Command::new("cargo");
+enum RequestedTarget {
```
djc, line 103: I think there's way too much stuff going on in this commit, and I'm not convinced all this stuff pulls its weight. For this PR, can we cut out this commit and/or minimize it to only build the unit test with cargo test?

src/bin/cargo-flamegraph.rs
```diff
@@ -270,32 +271,25 @@ fn find_unique_bin_target() -> anyhow::Result<BinaryTarget> {
...
-        [] => Err(anyhow!(
-            "crate has no binary targets: try passing `--example <example>` \
-                or similar to choose a binary"
-        )),
-        _ => Err(anyhow!(
-            "several possible targets found: {:?}, please pass `--bin <binary>` or \
-                `--example <example>` to cargo flamegraph to choose one of them",
-            bin_targets
-        )),
+        [] => Err(anyhow!("crate has no automatically selectable target: try passing `--example <example>` or similar to choose a binary")),
```
djc, line 337: I'd prefer that you don't change the formatting in the same commit as making code changes, as it makes it harder to review what you actually changed. In this case, I also prefer the previous formatting (that is, break long literals apart so they don't pass 100 columns).

ede1998: ok, I reverted the formatting and only adjusted the text.

src/bin/cargo-flamegraph.rs
ede1998: I agree that the commit is too big. I created a new [branch](https://github.com/ede1998/flamegraph/commits/build-with-cargo-test) where I kept the changes to a minimum. Commit 63d9d3be0e0f479e399d9362c1f4cf097b2b21a5 introduces building unit tests with `cargo test`. However this unfortunately means that we sometimes execute `cargo metadata` twice: once when finding a unit test target and again when determining whether to call `cargo test --lib` or `cargo test --bin <bin>`. As the first call occurs in `find_unique_target`, the most elegant way to avoid that seems to be to add the `kind` to the return value `BinaryTarget`, which is what I was trying to do in the mega commit. Though I can probably cut quite a bit of stuff out now that I have a look at it again.

djc: I'd be fine with calling `cargo metadata` twice for now and then separately following up on improvements to that.

src/bin/cargo-flamegraph.rs
djc: Thanks!

src/bin/cargo-flamegraph.rs
ede1998: Ok, I tried a third implementation in [this branch](https://github.com/ede1998/flamegraph/tree/parse-build-output-2). Commit dad539e9fe372466822dc5d59488067601b1a441 is another minimal implementation that builds unit test with `cargo test`. This one avoids running `cargo metadata` twice though but introduces an `kind` parameter to the build method which is only used when building a unit test target. I tried passing the `BinaryTarget in instead but that seems to massively bloat the commit again. So there is 3 versions now:
- Hopelessly bloated commit 346ed3281c4fb48d4c1ce2fe1711679b4a8371f8 in branch [ede1998:profile_unit_tests](https://github.com/ede1998/flamegraph/tree/profile_unit_tests)
- Running `cargo metadata` twice commit 63d9d3be0e0f479e399d9362c1f4cf097b2b21a5 in branch [ede1998:build-with-cargo-test](https://github.com/ede1998/flamegraph/tree/build-with-cargo-test)
- Extra `kind` parameter in build commit dad539e9fe372466822dc5d59488067601b1a441 in branch [ede1998:parse-build-output-2](https://github.com/ede1998/flamegraph/tree/parse-build-output-2) (whoops, that's not the branch name i was going for but it doesn't matter)

Which version do you like best?

djc: The extra `kind` parameter version looks good to me! Thanks for going through this comparison!

ede1998: Ok, I moved the commits into this PR and rebased them onto main.
Additionally, I made a small change to commit 707121d790b5ffad348f3c92c71e15a70536bd09: `trailing_arguments` now has a doc string and I updated the USAGE section in README again, which fixed two typos: `performace -> performance` and `default 0.1 -> 0.01`.

djc:
Thank you for all your work on this!

Issue #132, opened 2021-04-20
Fails to locate "integration" tests


bjchambers:
https://github.com/flamegraph-rs/flamegraph/blob/master/src/bin/cargo-flamegraph.rs#L321

When I run a simple example, it seems like the tests show up in `target/debug/deps` instead of `target/debug`.

As far as I can tell:

* `target/debug/<package>` is used for the collected tests within the `src` folder.
* `target/debug/deps/<file>` is used for `file.rs` in the `tests` folder.

If you have "integration" tests (in `tests` instead of `src`) you can't use `cargo flamegraph` to run them.

It looks for benchmarks the `deps` directory is added, but for tests, it seems like it may be a bit tricky since the directory seems to depend on the kind of test.

djc:
Maybe we could try both paths for tests and see which one exists? Should probably try the current approach first to make it more compatible for existing users. If you want to attempt a PR, that would be great!

Issue #79, opened 2020-03-18
flamegraph sometimes runs the wrong binary in `--test` mode


rcoh:
Command:
`cargo flamegraph --test fill_integ`

Binary that gets run (snooped via strace):
`rust-wasm/target/release/fill_integ-0d31b301ef6d061y`

contents of folder:
```
➜  rust-wasm git:(rust-wasm) ✗ ls -latr target/release/fill_integ-*
-rw-rw-r-- 1 russell russell      584 Mar 15 23:08 target/release/fill_integ-0d31b301ef6d0617.d
-rwxrwxr-x 3 russell russell 11542904 Mar 16 01:03 target/release/fill_integ-0d31b301ef6d0617
-rw-rw-r-- 1 russell russell      584 Mar 16 23:45 target/release/fill_integ-dc6d5509fba706c0.d
-rwxrwxr-x 2 russell russell 12489304 Mar 18 11:25 target/release/fill_integ-dc6d5509fba706c0
```

`-dc6....` is the new binary that should be run but flamegraph is runnin g `0d31b`.

Workaround is deleting the old binary.

djc:
I think #141 has fixed this, so I'll close this issue. If you want to give it a shot (from Git main), that would be great. If there's still an issue, please feel free to reopen.

PR #141, opened 2021-08-13
Run cargo build and read JSON output

ede1998:
## About
This PR changes `cargo flamegraph` so that it reads the output of `cargo build` to determine the patch of the binary that should be profiled.
Because of this change the automatic target selection had to be redone as previously it just chose the only binary from the target folder. Now, we run `cargo metadata`, find the only binary target of the workspace first and pass this target explictly. If there are multiple or no binary targets, we abort with an error.
I tested the new version several different dummy crates and workspaces as well as `ripgrep`. It seems to work fine.

## Relevant issue
- #139 This is where the idea came up.
- #132  This should be fixed by this PR assuming `cargo build` tells us the correct binary.
- #79 This should also be fixed.

## Follow-up PR
- #142

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -1,6 +1,6 @@
-use std::fs;
 use std::path::{Path, PathBuf};
 
+use cargo_metadata::{Artifact, Message};
 use structopt::StructOpt;
 
 use flamegraph::Workload;
@@ -101,6 +101,36 @@ struct Opt {
     trailing_arguments: Vec<String>,
 }
 
+impl Opt {
+    fn has_explicit_target(&self) -> bool {
+        self.bin.is_some() || self.bench.is_some() || self.example.is_some() || self.test.is_some()
+    }
+
+    fn target_kind(&self) -> &'static str {
+        match self {
+            Opt { bin: Some(_), .. } => "bin",
+            Opt {
+                example: Some(_), ..
+            } => "example",
+            Opt { test: Some(_), .. } => "test",
+            Opt { bench: Some(_), .. } => "bench",
+            _ => panic!("No target for profiling."),
+        }
+    }
+
+    fn target_name(&self) -> &str {
+        match self {
+            Opt { bin: Some(t), .. } => t,
+            Opt {
+                example: Some(t), ..
+            } => t,
+            Opt { test: Some(t), .. } => t,
+            Opt { bench: Some(t), .. } => t,
+            _ => panic!("No target for profiling."),
+        }
+    }
+}
+
 #[derive(Debug, StructOpt)]
 #[structopt(
     name = "cargo-flamegraph",
@@ -111,8 +141,9 @@ enum Opts {
     Flamegraph(Opt),
 }
 
-fn build(opt: &Opt) {
-    let mut cmd = std::process::Command::new("cargo");
+fn build(opt: &Opt) -> Vec<Artifact> {
+    use std::process::{Command, Output, Stdio};
+    let mut cmd = Command::new("cargo");
 
     // This will build benchmarks with the `bench` profile. This is needed
     // because the `--profile` argument for `cargo build` is unstable.
@@ -166,25 +197,29 @@ fn build(opt: &Opt) {
         cmd.arg("--no-default-features");
     }
 
+    cmd.arg("--message-format=json-render-diagnostics");
+
     if opt.verbose {
         println!("build command: {:?}", cmd);
     }
 
-    let mut child = cmd.spawn().expect("failed to spawn cargo build command");
+    let Output { status, stdout, .. } = cmd
+        .stderr(Stdio::inherit())
+        .output()
+        .expect("failed to execute cargo build command");
 
-    let exit_status = child
-        .wait()
-        .expect("failed to wait for cargo build child to finish");
+    let artifacts: Vec<_> = Message::parse_stream(&*stdout)
+        .filter_map(|m| match m {
+            Ok(Message::CompilerArtifact(artifact)) => Some(artifact),
+            Ok(_) => None,
+            Err(e) => {
+                panic!("failed to parse cargo build output: {:?}", e);
+            }
+        })
+        .collect();
 
     if !opt.dev {
-        cmd.arg("--message-format=json");
-
-        let output: Vec<u8> = cmd
-            .output()
-            .expect("failed to execute cargo build command")
-            .stdout;
-
-        let messages = cargo_metadata::Message::parse_stream(&*output);
+        let messages = Message::parse_stream(&*stdout);
 
         let mut has_debuginfo = false;
 
@@ -195,7 +230,7 @@ fn build(opt: &Opt) {
         };
 
         // get names of binaries in the workload
-        let workload_filenames = workload(opt)
+        let workload_filenames = workload(opt, &artifacts)
             .iter()
             .filter_map(|w| {
                 PathBuf::from(w)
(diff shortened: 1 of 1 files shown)


djc:
Hey, this is looking great. I'm on vacation so won't be able to look at this for a while, but I'll make sure to get back to it in a few weeks.

ede1998:
Thanks for the heads-up. Have a great vacation.

src/bin/cargo-flamegraph.rs
```diff
@@ -174,91 +176,86 @@ fn build(opt: &Opt) {
...
+        std::process::exit(1);
+    }
 
-        let messages =
-            cargo_metadata::Message::parse_stream(&*output);
+    artifacts
+}
 
-        let mut has_debuginfo = false;
+fn check_debug_info(opt: &Opt, artifacts: &[Artifact]) {
```
djc: Let's not create a single-use function here, and keep the checking of debug info inside `build()` as before. In general, this kind of refactoring should also go in a separate commit to make it easier to review.

src/bin/cargo-flamegraph.rs
```diff
@@ -174,91 +176,86 @@ fn build(opt: &Opt) {
...
+    let Output {
+        status,
+        stdout,
+        stderr,
+    } = cmd
+        .output()
+        .expect("failed to execute cargo build command");
+
+    let messages = Message::parse_stream(&*stdout);
+    let artifacts: Vec<_> = messages
```
djc, line 225: Is there a reason to bind `messages` here? It looks like we can call `filter_map()` directly on the return value of `Message:parse_stream()`.

src/bin/cargo-flamegraph.rs
```diff
@@ -203,54 +211,30 @@ fn build(opt: &Opt) -> Vec<Artifact> {
...
-    if !opt.dev {
-        check_debug_info(opt, &artifacts);
-    }
-
     if !status.success() {
-        eprintln!("cargo build failed: {:?}", stderr);
+        eprintln!(
+            "cargo build failed: {}",
+            String::from_utf8_lossy(&stderr)
+        );
```
djc: This should live in the previous commit.

src/bin/cargo-flamegraph.rs
```diff
@@ -203,54 +211,30 @@ fn build(opt: &Opt) -> Vec<Artifact> {
...
-            artifact.profile.debuginfo.unwrap_or(0) != 0
-                && workload_filenames.iter().rev().any(
-                    |w| {
-                        w.starts_with(&artifact.target.name)
-                    },
-                );
-    }
+fn check_debug_info(opt: &Opt, artifact: &Artifact) {
+    const NONE: u32 = 0;
+    let debug = artifact.profile.debuginfo.unwrap_or(NONE);
```
djc, line 247: These changes seem to be unrelated to the main goal of this commit?

src/bin/cargo-flamegraph.rs
```diff
@@ -259,109 +243,63 @@ fn check_debug_info(opt: &Opt, artifacts: &[Artifact]) {
...
-    }
-
-    let targets: Vec<String> = metadata
-        .packages
-        .into_iter()
-        .flat_map(|p| p.targets)
-        .filter(|t| t.crate_types.contains(&"bin".into()))
-        .map(|t| t.name)
-        .collect();
+fn select_executable(
```
djc, line 271: Prefer to inline this single-use function, as well.

src/bin/cargo-flamegraph.rs
```diff
@@ -306,11 +306,47 @@ fn workload(
...
 }
 
+fn find_unique_bin_target() -> String {
+    let mut metadata_command = MetadataCommand::new();
+    metadata_command.no_deps();
+    let metadata = metadata_command
+        .exec()
+        .expect("failed to access crate metadata");
+
+    let bin_targets: Vec<String> = metadata
```
djc: Let's turn this into a single call chain?

src/bin/cargo-flamegraph.rs
```diff
@@ -306,11 +306,47 @@ fn workload(
...
+        .packages
+        .into_iter()
+        .flat_map(|p| p.targets.into_iter())
+        .filter(|t| t.kind.contains(&"bin".into()))
+        .map(|t| t.name)
```
djc: We should use `filter_map()` instead of `filter().map()`. Ideally we should combine it with `flat_map()`, too, if at all possible.

src/bin/cargo-flamegraph.rs
```diff
@@ -306,11 +306,47 @@ fn workload(
     result
 }
 
+fn find_unique_bin_target() -> String {
```
djc: Let's also inline this single-use function.

src/bin/cargo-flamegraph.rs
```diff
@@ -153,7 +153,7 @@ enum Opts {
 }
 
 fn build(opt: &Opt) -> Vec<Artifact> {
-    use std::process::{Command, Output};
+    use std::process::{Command, Output, Stdio};
```
djc, line 158: Changes from this commit should be inlined into the previous commit.

src/bin/cargo-flamegraph.rs
```diff
@@ -258,6 +258,11 @@ fn check_debug_info(opt: &Opt, artifact: &Artifact) {
         );
         eprintln!("[profile.{}]", profile);
         eprintln!("debug = true\n");
+        eprintln!("Or set this environment variable:\n");
+        eprintln!(
+            "CARGO_PROFILE_{}_DEBUG=true\n",
+            profile.to_uppercase()
+        );
```
djc, line 267: Inline changes from this commit into the relevant previous commits, too, please.

src/bin/cargo-flamegraph.rs
```diff
@@ -320,19 +322,46 @@ fn workload(
...
-    let bin_targets: Vec<String> = metadata
+    let bin_targets: Vec<BinaryTarget> = metadata
         .packages
         .into_iter()
-        .flat_map(|p| p.targets.into_iter())
-        .filter(|t| t.kind.contains(&"bin".into()))
-        .map(|t| t.name)
+        .flat_map(|p| {
+            let Package { targets, name, .. } = p;
+            targets
```
djc, line 356: Why can't we use `p.targets` directly anymore, as before?

djc (commented):
So here's an initial pass. As is, I found this pretty challenging to review, please follow up on these comments and I'll do another pass. (If you haven't adjusted commit history before, I usually use interactive rebases via `git rebase -i`, where you can use the `squash` and `fixup` commands to fold newer commits into older ones.)

src/bin/cargo-flamegraph.rs
ede1998: No because `p` is used in the closure again and the borrow checker does not like that:
```
error[E0382]: use of partially moved value: `p`
   --> src/bin/cargo-flamegraph.rs:356:22
    |
355 |                 .into_iter()
    |                  ----------- `p.targets` partially moved due to this method call
356 |                 .map(move |t| (p.name.clone(), t))
    |                      ^^^^^^^^  ------ use occurs due to use in closure
    |                      |
    |                      value used here after partial move
    |
note: this function takes ownership of the receiver `self`, which moves `p.targets`
```

src/bin/cargo-flamegraph.rs
ede1998: No. I think I originally did this because the code didn't look good after `cargo fmt` because of the 60 characters max width in `.rustfmt.toml`. But it seems like I changed the code again so it looks just fine now.

src/bin/cargo-flamegraph.rs
ede1998: Sure I can inline them but it is quite long. That's why I split it off into another function to keep it more readable. Because of the function name, you can immediately see what it does which is not be the case if I'll inline the block (unless I add a describing comment).
The same goes for the other single use functions.

src/bin/cargo-flamegraph.rs
ede1998: I combined it to `filter_map` but I didn't see a way to combine it with `flat_map`.

src/bin/cargo-flamegraph.rs
ede1998: I inlined all the other functions you suggested but I didn't inline this one yet. If you insist, I can inline it but I think `main` will be too long then.

ede1998:
@djc Ok, I fixed up the commit history like you suggested and also implemented (most of) the other suggestions. I had to force-push to get the new history up here but I think that was to be expected. It seems though that the CI-workflow didn't automatically start again and I don't have permission to do that:
`First-time contributors need a maintainer to approve running workflows.`

src/bin/cargo-flamegraph.rs
```diff
@@ -174,28 +208,31 @@ fn build(opt: &Opt) {
...
-        .expect("failed to spawn cargo build command");
+    let Output { status, stdout, .. } = cmd
+        .stderr(Stdio::inherit())
+        .output()
+        .expect("failed to execute cargo build command");
 
-    let exit_status = child.wait().expect(
-        "failed to wait for cargo build child to finish",
-    );
+    let artifacts: Vec<_> = Message::parse_stream(&*stdout)
```
djc, line 224: Ah, I missed that this now duplicates the `Message::parse_stream()` call. In that case, it still makes sense to factor it out so that `!opt.dev` can share the same result.

src/bin/cargo-flamegraph.rs
```diff
@@ -174,28 +208,31 @@ fn build(opt: &Opt) {
...
+        .filter_map(|m| match m {
+            Ok(Message::CompilerArtifact(artifact)) => {
```
djc, line 226: This might be easier to folllow if you use `m.map()` and `collect()` into a `Result`, which you can then just expect on the outside -- maybe try it and see which one you think is easier to follow?

src/bin/cargo-flamegraph.rs
```diff
@@ -256,116 +293,59 @@ fn build(opt: &Opt) {
...
-        } else {
+        // target.kind is an array for some reason. No idea why though, it always seems to contain exactly one element.
+        // If you know why, feel free to PR and handle kind properly.
+        let artifact = artifacts.iter().find(|a| {
+            a.target.name == target
+                && a.target.kind.contains(&kind)
+                && a.executable.is_some()
+        });
+
+        let artifact = artifact.unwrap_or_else(|| {
```
djc, line 266: Let's chain this directly? I think we can return `artifacts.iter().find(...).unwrap_or_else().executable.as_deref().expect()`? Also I'd probably leave that outer `expect()` to the caller.

src/bin/cargo-flamegraph.rs
```diff
@@ -337,6 +275,33 @@ fn workload(
...
+        const NONE: u32 = 0;
+        let debug_level =
+            artifact.profile.debuginfo.unwrap_or(NONE);
+
+        if !opt.dev && debug_level == NONE {
+            let profile = if opt.bench.is_some() {
+                "bench"
+            } else {
+                "release"
+            };
```
djc, line 289: This would be more succinctly expressed as a plain `match` expression.

src/bin/cargo-flamegraph.rs
```diff
@@ -316,11 +316,53 @@ fn workload(
...
+        })
+        .collect();
+
+    match bin_targets.as_slice() {
+        [target] => {
+            eprintln!(
+                "automatically selected {} as it is the only binary target",
+                target
+            );
+            target.clone()
```
djc, line 371: I think we can return `bin_targets[0]` here instead of cloning?

src/bin/cargo-flamegraph.rs
```diff
@@ -316,11 +316,53 @@ fn workload(
...
+fn find_unique_bin_target() -> String {
+    let metadata = MetadataCommand::new()
+        .no_deps()
+        .exec()
+        .expect("failed to access crate metadata");
+
+    let bin_targets: Vec<String> = metadata
+        .packages
+        .into_iter()
+        .flat_map(|p| p.targets.into_iter())
```
djc: Any reason not to inline the `filter_map()` into the `flat_map()` closure here?

src/bin/cargo-flamegraph.rs
```diff
@@ -316,11 +316,53 @@ fn workload(
     result
 }
 
+fn find_unique_bin_target() -> String {
+    let metadata = MetadataCommand::new()
+        .no_deps()
+        .exec()
+        .expect("failed to access crate metadata");
+
+    let bin_targets: Vec<String> = metadata
```
djc: I think we can chain this stuff together?

djc:
Sorry, I generated a bunch of conflicts here due to reformatting with the default line length. You should be able to just keep all your changes and reformat after merge/rebase.

src/bin/cargo-flamegraph.rs
ede1998: No:
```
Error[E0507]: cannot move out of index of `Vec<BinaryTarget>`
   --> src/bin/cargo-flamegraph.rs:367:13
    |
367 |             bin_targets[0]
    |             ^^^^^^^^^^^^^^ move occurs because value has type `BinaryTarget`, which does not implement the `Copy` trait
```

I managed to make it work but I'm not convinced that's better:
```rust
match bin_targets.as_slice() {
    [_] => bin_targets.into_iter().next().unwrap(),
...
```

While developing I also tried to destructure the `Vec` itself but that also leads to borrow checker issues and unsized locals for the catch-all arm.

ede1998: Ok, just had another look into `Vec` docs. `bin_targets.remove(0);`works as well if I make `bin_targets` mutable. I think I will go with that.

src/bin/cargo-flamegraph.rs
ede1998: I'd say it's a bit harder to read. But maybe that's just because I have to get used to it. While I like the `expect` on the outside, the `match` and `transpose` combo is a bit complex to read in my opinion. If you have any suggestions to simplify this, that would be great. The only way I see right now is to add an extension trait for `Message` that provides a `fn into_artifact(self) -> Option<Artifact>` but that seems like too much.

```rust
let artifacts = Message::parse_stream(&*stdout)
    .filter_map(|r| {
        r.map(|m| match m {
            Message::CompilerArtifact(a) => Some(a),
            _ => None,
        })
        .transpose()
    })
    .collect::<Result<Vec<_>, _>>()
    .expect("failed to parse cargo build output");
```

src/bin/cargo-flamegraph.rs
ede1998: I'll leave that as is because the second `Message::parse_stream` is removed in this PR anyway.

src/bin/cargo-flamegraph.rs
ede1998: > I think we can return ...?

In this commit, yes but in the next commit 010ad1a8a6d3b04dce6dc49ada8e1641d669e282 I'm adding the debug info check inbetween `unwrap_or_else` and `executable`. This check needs the artifact.

> Also I'd probably leave that outer expect() to the caller.

It's not a function anymore just a block expression. Also the expect itself should never ever occur unless someone changes this part of the code again because the `find` predicate only allows artifacts with executables.
```rust
let binary_path = {
    let artifact = artifacts.iter().find(...).unwrap_or_else(...);
    artifact.executable.as_deref().expect(...)
};
```

ede1998:
Ok, I implemented your suggestions or added a comment.

src/bin/cargo-flamegraph.rs
djc: Ah, that's fine then.

src/bin/cargo-flamegraph.rs
djc: Can we express that in some way that we don't have the unnecessary `expect()` call, then? I guess we could `find_map()` in a way that we return the deref of executable directly along with the `artifact`?

src/bin/cargo-flamegraph.rs
djc: Yup, that sounds good!

djc:
I'd like to have the reformatting applied in the earlier commits. This should be easy to automate with `git rebase -x "cargo fmt && git commit --amend -a"` or something like that (but if it's too much pain for you I'd be happy to do it myself).

src/bin/cargo-flamegraph.rs
djc: That's fair -- go with the approach you like best.

src/bin/cargo-flamegraph.rs
ede1998: Excellent suggestion. I like the new version much more.

ede1998:
> I'd like to have the reformatting applied in the earlier commits.

Ok I misunderstood your previous comment then. I fixed the formatting again and implemented the new suggestions.

djc:
Thanks for all your work on this, this is a great improvement!

PR #142, opened 2021-08-20
Combine #140 and #141, fix merge conflicts and add improvements

ede1998:
## About

This PR contains the contents of #140 and #141 merged together into one PR with all the merge conflicts resolved. In addition, it addresses the following issues from #140:
- `test --no-run` should be used instead of `build --tests`; for this I also took the liberty to refactor the build method a bit.
- Wrong target binary selection; pretty much fixed itself by having #141 in the branch.

 I tested the new version with several different dummy crates and workspaces as well as ripgrep. It seems to work fine. (My tests only entailed `cargo-flamegraph` itself though, not the machinery behind it that is required to generate a (correct) flamegraph.)


djc:
Since the changes from #141 have been merged, I'm going to close this is in favor of #140.

PR #152, opened 2021-09-02
fix target selection

ede1998:
@djc Found a bug that always enabled automatic target selection.


djc:
Oops, thanks for catching that!

PR #156, opened 2021-09-29
remove hint about cargo flamegraph not picking up debug info from README

ede1998:
The finicky code was replaced in commit https://github.com/flamegraph-rs/flamegraph/pull/141/commits/53e0cc456146aa6a82ea6fb57db7af349c549c3e

I tried it out and the warning disappears now when building with debug info.


djc:
Thanks!

PR #157, opened 2021-10-04
lld on linux requires a special flag to make flamegraph work

mahkoh:


mstange:
Do we know why this is?

mahkoh:
https://bugs.chromium.org/p/chromium/issues/detail?id=919499#c16

mahkoh:
Someone might also want to note this issue in https://github.com/rust-lang/rust/issues/71515. From a quick look https://github.com/rust-lang/rust/issues/79555 might also be related. Unfortunately I'm banned from the rust-lang org so I cannot do it myself.

README.md
```diff
@@ -15,6 +15,14 @@ flamegraph generation library!
 
 Windows is getting [dtrace support](https://techcommunity.microsoft.com/t5/Windows-Kernel-Internals/DTrace-on-Windows/ba-p/362902), so if you try this out please let us know how it goes :D
 
+**Note**: If you're using lld on on Linux, you must use the `--no-rosegment` flag. Otherwise perf will not be able to generate accurate stack traces. For example
```
djc, line 18:
```suggestion
**Note**: If you're using lld on Linux, you must use the `--no-rosegment` flag. Otherwise perf will not be able to generate accurate stack traces. For example
```

Maybe link to the Chromium issue here, as well?

mahkoh: Done

mahkoh:
Maybe because I used the github UI to edit the file.

PR #169, opened 2021-11-08
fix #168: get metadata from correct crate when using manifest-path

### LINKED ISSUE #168: `--manifest-path` doesn't seem to work (or behaviour does not match usual semantics)
Normally `--manifest-path` allows cargo-ing within a specific project from an unrelated location e.g. `cargo run --manifest-path ~/some/project/Cargo.toml` will run `~/some/project` but with the current `PWD`, which is useful for testing CLI utilities which work on `PWD` (e.g. because an override seems unnecessary or hasn't been built yet).

Doing that with flamegraph, however, yields the following:
```
Error: failed to access crate metadata

Caused by:
    `cargo metadata` exited with an error: error: could not find `Cargo.toml` in `/current/working/directory` or any parent directory
```

masklinn:
In the implementation of #121, `--manifest-path` is forwarded to
`cargo build` so the correct crate gets built and profiled, however it
is not forwarded to `cargo metadata`.

This could impact target inference, but more importantly prevented
`--manifest-path` from being used in order to profile a crate without
being in any crate root e.g. if the crate being profile works off of
`$PWD`, `cargo-flamegraph` was not able to run if said `PWD` was not
in a crate context.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Context};
 use cargo_metadata::{Artifact, Message, MetadataCommand, Package};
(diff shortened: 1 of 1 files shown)


masklinn:
@ede1998

djc (approved):

masklinn:
@ede1998 well apparently I missed the introduction of `Option::as_deref` and didn't even wonder about its existence on this very same line. Thanks for the introduction that'll come in handy! (and it should be fixed).

ede1998 (approved):

djc:
Thanks to both of you!

PR #170, opened 2021-11-08
fix #31: mention macOS status explicitely

### LINKED ISSUE #31: docs: README should mention MacOS support
Whether flamegraph supports MacOS or not, the README should say so.

masklinn:
Disabling System Integrity Protection only seems necessary to
trace *system utilities*: binaries living in `/System`, `/usr` (except
`/usr/local`), `/bin`, `/sbin`, `/var`, and preinstalled applications.

Therefore tracing a rust utility probably should not conflict with
SIP, but it still requires running a superuser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,15 @@ be acceptable for your security needs etc...
 echo -1 | sudo tee /proc/sys/kernel/perf_event_paranoid
 ```
 
+### DTrace on macOS
+
+On macOS, there is no alternative to running as superuser in order to
+enable dtrace. The simplest way is to use `--root`, this way `rustc`
+will be run normally but the binary will get run as superuser.
+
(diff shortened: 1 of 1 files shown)


masklinn:
Note: did this on a rather outdated macOS 10.11 and it works fine (running `cargo flamegraph --root` is how I hit #168 — and worked around it), however if someone has access to a macOS 11 or 12 machine to make sure that'd be nice.

Given how recent the latest comments on #31 are maybe @capnfabs or @rsqv were / are running recent versions?

djc (approved):

fenhl:
I can confirm that these instructions work on macOS Monterey 12.0.1 with System Integrity Protection enabled.

djc:
@fenhl thanks for testing!

PR #172, opened 2021-11-16
Add `skip_after` option to flamegraph

### LINKED ISSUE #171: allow customization of CollapseOptions
inferno's `CollapseOptions` have some useful tunings. It could be nice if we could customize those via flamegraph. 
See https://github.com/flamegraph-rs/flamegraph/blob/main/src/lib.rs#L271  where they are instantiated.

1. Would you agree to this enhancement?
2. How would you suggest to propagate them? One way is to explicitly declare each option and manually propagate them, 
but I thought it could be nice to do something like `--collapse_options=key1=val1,key2=val2...` and let the underlying struct deserialize those key pairs. This approach would reduce the maintenance overhead and give 100% coverage of these options.

romange:
Fixes #171

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -268,18 +268,14 @@ pub fn generate_flamegraph_for_workload(
 
     let collapsed_writer = BufWriter::new(&mut collapsed);
 
-    let mut collapse_options: CollapseOptions;
+    #[allow(unused_mut)]
+    let mut collapse_options = CollapseOptions::default();
+
     #[cfg(target_os = "linux")]
     {
-        collapse_options = CollapseOptions::default();
         if let Some(skip_after) = &opts.flamegraph_options.skip_after {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -108,6 +108,15 @@ mod arch {
             .context("unable to call perf script")?
             .stdout)
     }
+
+    pub fn set_collapse_options(
```
djc, line 112: I'd probably prefer not to have this as a separate method -- I think we could just inline this code?

romange: it should be under cfg attribute conditional. How can I inline a milti-line block under `#cfg` attribute?

djc: `#[cfg()] { /* your code goes here */ }`

romange: fixed

djc (approved):

djc:
Thanks!

PR #174, opened 2021-11-20
fix: permission denied reading stacks file

### LINKED ISSUE #173: bug: permission denied reading stacks file with dtrace + root (macOS)
Hi there, awesome tool!

When running the latest commit on main (eccf867) with `--root` on macOS, on my system the generated `cargo-flamegraph.stacks` file is owned by `root` and the user does not have permission to read it:

```console
rustfmt % cargo flamegraph --root --bench test_corpus -- --bench tests/source/trait.rs
    Finished bench [optimized + debuginfo] target(s) in 0.15s
Password:
dtrace: system integrity protection is on, some features will not be available

dtrace: description 'profile-997 ' matched 1 probe

	<...snip...>

dtrace: pid 48896 has exited
Error: failed to open dtrace output file 'cargo-flamegraph.stacks'

Caused by:
    Permission denied (os error 13)
```

And the user has no read permission:

```console
rustfmt % ls -alh cargo-flamegraph.stacks
-rw------- 1 root staff 3.5K Nov 20 18:21 cargo-flamegraph.stacks
```

Truthfully I expect this would have been noticed already if it affected more people, so it may be something on my machine, though I don't believe I've set anything to do tighten up permissions... Either way, I think it's relatively easy to fix up 👍

domodwyer:
Ensures the `cargo-flamegraph.stacks` file generated when using the dtrace profiler can be read to be collapsed by setting the owner to `$USER` instead of `root` 👍 

Fixes #173.

---

* fix: set cargo-flamegraph.stacks file owner (ae0f5d9)

      This commit adds code to execute:

      	sudo chown $USER cargo-flamegraph.stacks

      When the following conditions are met:

      	* The flamegraph was generated with dtrace (aka. not on Linux)
      	* The user passed --root on the command line
      	* The USER env is set

      This ensures the cargo-flamegraph.stacks file wrote by dtrace as "root" is
      readable by the non-root user.

* refactor: fix clippy lint (e9164ca)

      Removes redundant 'static on const vars.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,7 +89,11 @@ mod arch {
         (command, perf_output)
     }
 
-    pub fn output(perf_output: Option<String>, script_no_inline: bool) -> anyhow::Result<Vec<u8>> {
+    pub fn output(
+        perf_output: Option<String>,
+        script_no_inline: bool,
+        _sudo: bool,
+    ) -> anyhow::Result<Vec<u8>> {
         let perf = env::var("PERF").unwrap_or_else(|_| "perf".to_string());
         let mut command = Command::new(perf);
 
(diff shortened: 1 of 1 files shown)


djc (commented):
Makes sense, thanks for fixing it up!

djc (approved):

domodwyer:
Thanks @djc! Any chance of getting this merged?

djc:
Yup, sorry for the delay. Thanks!

PR #177, opened 2021-12-01
Update installation instructions for Debian/Ubuntu

smu160:
Hi,

This PR is meant to address [issue 176](https://github.com/flamegraph-rs/flamegraph/issues/176).

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,16 @@ rustflags = ["-Clink-arg=-fuse-ld=lld", "-Clink-arg=-Wl,--no-rosegment"]
 
 ## Installation
 
-Requirements on Linux (Ubuntu/Debian):
-```
+Requirements on Linux:
+
+#### Debian
+```bash
 sudo apt install -y linux-perf
+```
+
+#### Ubuntu
+```bash
+sudo apt install linux-tools-common linux-tools-generic linux-tools-`uname -r`
 ```
 
 `flamegraph` not `cargo-flamegraph`! (`cargo-flamegraph` is an inactive crate as of March 2019)
--- a/README.md
+++ b/README.md
@@ -38,6 +38,11 @@ Not working on aarch, use a Debian distribution, or make a PR with your solution
 sudo apt install linux-tools-common linux-tools-generic linux-tools-`uname -r`
 ```
 
+#### Pop!_OS
+```bash
+sudo apt install linux-tools-common linux-tools-generic
+```
+
 `flamegraph` not `cargo-flamegraph`! (`cargo-flamegraph` is an inactive crate as of March 2019)
 
 ```
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ Requirements on Linux:
 sudo apt install -y linux-perf
 ```
 
-#### Ubuntu
+#### Ubuntu (x86)
+Not working on aarch, use a Debian distribution, or make a PR with your solution for Ubuntu
 ```bash
 sudo apt install linux-tools-common linux-tools-generic linux-tools-`uname -r`
 ```
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ rustflags = ["-Clink-arg=-fuse-ld=lld", "-Clink-arg=-Wl,--no-rosegment"]
 
 Requirements on Linux:
 
-#### Debian
+#### Debian (x86 and aarch)
 ```bash
 sudo apt install -y linux-perf
 ```


kraktus:
It does not work on ubuntu 20.04 (tested on docker)

smu160:
@kraktus 

Thank you for bringing this up. Are you able to get perf installed using just:

```
sudo apt install linux-tools-common linux-tools-generic
```

kraktus:
Unfortunately `perf` does not seem to be installed with these packages.
```
automatically selected target test-flamegraph-ubuntu-install in package test-flamegraph-ubuntu-install as it is the only valid target
    Finished release [optimized + debuginfo] target(s) in 0.00s
WARNING: perf not found for kernel 5.10.47

  You may need to install the following packages for this specific kernel:
    linux-tools-5.10.47-linuxkit
    linux-cloud-tools-5.10.47-linuxkit

  You may also want to install one of the following packages to keep up to date:
    linux-tools-linuxkit
    linux-cloud-tools-linuxkit
failed to sample program
```
`linux-tools-linuxkit`, ` linux-cloud-tools-linuxkit` are not available on Ubuntu either.

[Dockerfile used](https://gist.github.com/kraktus/4552ff9f1669ac5fcdecb8b6692fbea8).

Could be a docker artefact, I don't have a native access to Ubuntu to try the command added by the PR. Have you tested it natively?

smu160:
@kraktus 

This is interesting. I'll try it today on a VM instance of Ubuntu 20.04, as well as Debian 11.

Yes, I tested natively on Ubuntu 21.10.

Thank you!

kraktus:
For Debian 11, `apt install -y linux-perf` works on Docker, [that is what I ended up using](https://github.com/kraktus/rs-retroboard-chess/blob/8301795dd2a6348f20256b109ceed6cadffcad57/Dockerfile#L14).

smu160:
@kraktus 

##### Debian 11:

I was able to confirm that `sudo apt install linux-perf` works, on a VM instance.

##### Ubuntu 20.04

```
sudo apt install linux-tools-common linux-tools-generic
perf --version
```
gives us:
```
WARNING: perf not found for kernel 5.11.0-41

You may need to install the following packages for this specific kernel:
     linux-tools-5.11.0-41-generic
     linux-cloud-tools-5.11.0-41-generic

You may also want to install one of the following packages to keep up to date: 
    linux-tools-generic
    linux-cloud-tools-generic
```

The fix is: 

```bash
sudo apt install linux-tools-common linux-tools-generic linux-tools-`uname -r`
```

ZetaNumbers:
> sudo apt install linux-tools-common linux-tools-generic
> perf --version

I am on `Pop!_OS 21.04` and this doesn't give me any warnings. Non generic packages are installed automatically. Perhaps it was packages' bug?

kraktus:
> The fix is:
> 
> ```shell
> sudo apt install linux-tools-common linux-tools-generic linux-tools-`uname -r`
> ```


As pointed in my first message this does not work for me, because it seems that we don't have the same linux kernel version when testing
```bash
$cat /etc/os-release 
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
(...)
$uname -r
5.10.76-linuxkit
$uname -a
Linux f4711d79fe28 5.10.76-linuxkit #1 SMP PREEMPT Mon Nov 8 11:22:26 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux
```

I'm not well versed in linux kernel versions, but it's probably because I'm running on aarch64

smu160:
@kraktus 

Thank you for bringing this up. I'll look into whether there is a more robust solution.

kraktus:
For now would be already much better to fix it for x86 and list as caveats it’s not available for aarch

smu160:
> For now would be already much better to fix it for x86 and list as caveat it’s not available for aarch on ubuntu

@kraktus Apologies for the belated response. Did you mean that we should just list the known installation instructions for various versions of Ubuntu, Pop!_OS, etc., but make note of potential pitfalls for cases such as aarch?

Thank you!

README.md
```diff
@@ -25,11 +25,18 @@ rustflags = ["-Clink-arg=-fuse-ld=lld", "-Clink-arg=-Wl,--no-rosegment"]
...
-Requirements on Linux (Ubuntu/Debian):
-```
+Requirements on Linux:
+
+#### Debian
+```bash
 sudo apt install -y linux-perf
 ```
 
+#### Ubuntu
```
kraktus, line 35:
```suggestion
#### Ubuntu (x86)
Not working on aarch, use a Debian distribution, or make a PR with your solution for Ubuntu
```

README.md
```diff
@@ -25,11 +25,18 @@ rustflags = ["-Clink-arg=-fuse-ld=lld", "-Clink-arg=-Wl,--no-rosegment"]
 
 ## Installation
 
-Requirements on Linux (Ubuntu/Debian):
-```
+Requirements on Linux:
+
+#### Debian
```
kraktus, line 30:
```suggestion
#### Debian (x86 and aarch)
```

kraktus:
Yes 👍 
Requirements on Linux:

> > For now would be already much better to fix it for x86 and list as caveat it’s not available for aarch on ubuntu
> 
> @kraktus Apologies for the belated response. Did you mean that we should just list the known installation instructions for various versions of Ubuntu, Pop!_OS, etc., but make note of potential pitfalls for cases such as aarch?
> 
> Thank you!

Yes exactly 👍

djc (approved):
Thanks! Any further improvements are welcome as separate PRs.

PR #179, opened 2021-12-13
Follow-up fixes for #178 Add support for flamechart

ishitatsuyuki:
I missed several things in #178. This is a follow-up fix.


djc:
Thanks for cleaning things up, much appreciated!

PR #182, opened 2021-12-24
Fix #180: Handle workspaces more similar to `cargo build/run/test`

### LINKED ISSUE #180: When used in a workspace, behavior is different from other cargo subcommands
If you're inside a sub-dir of a workspace, most cargo commands such as "cargo test", "cargo build", "cargo check", "cargo bench" will only operate on any crates in or beneath the current directory. e.g. If you have a workspace with multiple crates and you `cd` into one of those crate directories, cargo will only build that crate, test that crate, bench that crate, etc. - unless you tell it otherwise with the "--workspace" flag.

But "cargo flamegraph" will attempt to compile the whole workspace.  It appears you can select an individual crate with "--package" but it would be better and more consistent UX for it to behave the same as other cargo commands.

ede1998:
## About
This PR fixes the reproducible issues mentioned in #180. Namely this is the different behavior between standard cargo commands and `cargo flamegraph` when using workspaces. The default cargo commands by default only take into account crates in the working directory while `cargo flamegraph` was working on the entire working space.

The PR tries to amend that by changing `cargo flamegraph` behavior in the following ways:
- `cargo flamegraph` chooses a target only from the current working directory unless `--package` or `--manifest-path` are used (instead of the entire workspace).
- `cargo flamegraph` suggests targets only from the current working directory unless `--package` or `--manifest-path` are used (instead of the entire workspace).

I tested the new version with several different dummy crates and workspaces as well as ripgrep. It seems to work fine and just like the standard cargo commands.

## Breaking changes (Not sure if this is relevant for anyone though.)
As described above, `cargo flamegraph` no longer works on an entire workspace but only on the crate in the current working directory. This means that calling `cargo flamegraph` in a crate within a workspace tries to execute the binary within this crate instead of the workspace binary.

Example: calling `cargo flamegraph` in `ripgrep/crates/printer` (a library crate) now errors because it cannot find a binary instead of executing `rg` (the workspace executable).

## Relevant issues
- #180

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -269,6 +269,42 @@ impl std::fmt::Display for BinaryTarget {
     }
 }
 
+pub fn find_crate_root(manifest_path: Option<&Path>) -> anyhow::Result<PathBuf> {
+    match manifest_path {
+        Some(path) => {
+            let path = path.parent().ok_or_else(|| {
+                anyhow!(
+                    "the manifest path '{}' must point to a Cargo.toml file",
+                    path.display()
+                )
+            })?;
+
+            path.canonicalize().with_context(|| {
+                anyhow!(
+                    "failed to canonicalize manifest parent directory '{}'\nHint: make sure your manifest path is exists and points to a Cargo.toml file",
+                    path.display()
+                )
+            })
+        }
+        None => {
+            let cargo_toml = "Cargo.toml";
+            let cwd = std::env::current_dir().context("failed to determine working directory")?;
+
+            for current in cwd.ancestors() {
+                if current.join(cargo_toml).exists() {
+                    return Ok(current.to_path_buf());
+                }
+            }
+
+            Err(anyhow!(
+                "could not find '{}' in '{}' or any parent directory",
+                cargo_toml,
+                cwd.display()
+            ))
+        }
+    }
+}
+
 fn find_unique_target(
     kind: &[&str],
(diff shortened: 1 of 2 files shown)


src/bin/cargo-flamegraph.rs
```diff
@@ -269,6 +269,58 @@ impl std::fmt::Display for BinaryTarget {
...
+                anyhow!(
+                    "failed to canonicalize manifest parent directory '{}'\nHint: make sure your manifest path is exists and points to a Cargo.toml file",
+                    path.display()
+                )
+            })
+        }
+        None => {
+            // heavily inspired by https://github.com/rust-lang/cargo/blob/1f12b88497f96b67ffdd7b1d41df98f690091358/src/cargo/util/important_paths.rs#L6
+            let valid_cargo_toml_file_name = "Cargo.toml";
+            let invalid_cargo_toml_file_name = "cargo.toml";
```
djc, line 292: While it makes sense for Cargo to do this, I think we can skip the details of checking for `cargo.toml`?

djc (commented):
This mostly looks good to me, thanks for improving the situation here!

src/bin/cargo-flamegraph.rs
ede1998: Sure, fixed it. I also removed the comment as the new version is now a pretty naive approach without much resemblance to the original source.

djc:
Awesome, thanks!

PR #183, opened 2022-01-03
Update arg parsing to clap v3

tranzystorek-io:
Slightly breaking change: trailing arguments are now marked as `last` i.e. they have to be put after a `--` marker, which seems to be more manageable/reasonable.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -2,103 +2,62 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Context};
 use cargo_metadata::{Artifact, Message, MetadataCommand, Package};
-use structopt::StructOpt;
+use clap::Parser;
 
 use flamegraph::Workload;
 
-#[derive(Debug, StructOpt)]
-#[structopt(
-    setting = structopt::clap::AppSettings::TrailingVarArg
-)]
+/// A cargo subcommand for generating flamegraphs, using inferno
+#[derive(Parser, Debug)]
+#[clap(name = "cargo-flamegraph", version)]
 struct Opt {
     /// Build with the dev profile
-    #[structopt(long = "dev")]
+    #[clap(long)]
     dev: bool,
 
     /// package with the binary to run
-    #[structopt(short = "p", long = "package")]
+    #[clap(short, long)]
     package: Option<String>,
 
     /// Binary to run
-    #[structopt(
-        short = "b",
-        long = "bin",
-        conflicts_with = "bench",
-        conflicts_with = "unit-test",
-        conflicts_with = "example",
-        conflicts_with = "test"
-    )]
+    #[clap(short, long, group = "exec-args")]
     bin: Option<String>,
 
     /// Example to run
-    #[structopt(
-        long = "example",
-        conflicts_with = "bench",
-        conflicts_with = "unit-test",
-        conflicts_with = "bin",
-        conflicts_with = "test"
-    )]
+    #[clap(long, group = "exec-args")]
     example: Option<String>,
 
     /// Test binary to run (currently profiles the test harness and all tests in the binary)
-    #[structopt(
-        long = "test",
-        conflicts_with = "bench",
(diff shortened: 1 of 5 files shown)


src/bin/cargo-flamegraph.rs
```diff
@@ -120,44 +79,44 @@ fn build(opt: &Opt, kind: impl IntoIterator<Item = String>) -> anyhow::Result<Ve
         cmd.arg("--release");
     }
 
-    if let Some(ref package) = opt.package {
+    if let Some(package) = &opt.package {
```
djc, line 82: These changes seem unrelated (that is, unnecessary for porting to clap). Mind putting them in a separate PR or commit?

src/bin/flamegraph.rs
```diff
@@ -1,47 +1,43 @@
...
 struct Opt {
     /// Profile a running process by pid
-    #[structopt(short = "p", long = "pid")]
+    #[clap(short, long)]
     pid: Option<u32>,
 
     /// Generate shell completions for the given shell.
-    #[structopt(long = "completions", conflicts_with = "pid")]
-    completions: Option<structopt::clap::Shell>,
+    #[clap(long, value_name = "SHELL", conflicts_with_all = &["pid", "trailing-arguments"])]
```
djc, line 17: Why spell out the `value_name` here?

src/bin/flamegraph.rs
```diff
@@ -1,47 +1,43 @@
...
+    completions: Option<Shell>,
 
-    #[structopt(flatten)]
+    #[clap(flatten)]
     graph: flamegraph::Options,
 
-    #[structopt(parse(from_os_str), long = "perfdata", conflicts_with = "pid")]
+    #[clap(long = "perfdata", conflicts_with = "pid", parse(from_os_str))]
```
djc, line 23: Is reordering necessary here or just a style tweak?

djc (commented):
Thanks for working on this! In order to make this easy to review, I'd like to have just the minimal changes required to get it compiling with clap, and then have a follow-up PR (or follow-up commits in this PR) to further improve the command-line parsing.

Also could you please elaborate with some examples on things that are now allowed or are no longer allowed with regards to the argument parsing?

src/bin/flamegraph.rs
tranzystorek-io: It describes what we're actually specifying, i.e. the target shell, otherwise clap just copies the field name

src/bin/flamegraph.rs
tranzystorek-io: The latter

tranzystorek-io:
Thank you for having me! 😄

As for the new things, clap 3 mostly just cleans up all the APIs and incorporates old structopt under the `derive` umbrella.

Among the more interesting new additions is the `ArgEnum` trait which simplifies using enums as arguments, e.g.:

```rust
use clap::ArgEnum;

#[derive(ArgEnum)]
enum ProcessingMode {
    Batch,
    Interactive,
}

#[derive(Parser)]
struct Cli {
    #[long, arg_enum]
    mode: ProcessingMode,
}
```

This would accept args like `myprogram --mode batch` without any manual impls like `FromStr`, as well as display possible variants in the help message. The variant names can also be switched similarly to serde, to `lowercase`, `snake_case`, `kebab-case` etc.

tranzystorek-io:
I went ahead and disabled the CLI portion of `inferno` since it introduced unused clap 2 dependency

tranzystorek-io:
> Also could you please elaborate with some examples on things that are now allowed or are no longer allowed with regards to the argument parsing?

I seem to have missed the point of your question, I'm assuming you were asking about the breaking changes I introduced.

Now that the `trailing_arguments` in both `flamegraph` and `cargo_flamegraph` are marked as `last`, they **must** be placed after the `--`, e.g.:

```console
cargo flamegraph -F 10000 -- --my --args -a -b -c
```

This is slightly more restrictive, but also now the user unambiguously knows how to pass any arguments to their own application.

djc:
Thanks!

PR #195, opened 2022-03-12
Document --skip-after, update inferno (skip_after changed)

martinxyz:
See: https://github.com/jonhoo/inferno/pull/231

We also should bump the version to 0.6.0 because the FlamegraphOptions is public, so this is a breaking change. Not sure if anyone is using this API, though. I did not include a version bump in this PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -293,9 +293,7 @@ pub fn generate_flamegraph_for_workload(workload: Workload, opts: Options) -> an
 
     #[cfg(target_os = "linux")]
     {
-        if let Some(skip_after) = &opts.flamegraph_options.skip_after {
-            collapse_options.skip_after = Some(skip_after.into())
-        }
+        collapse_options.skip_after = opts.flamegraph_options.skip_after.clone();
     }
 
     Folder::from(collapse_options)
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -293,9 +293,7 @@ pub fn generate_flamegraph_for_workload(workload: Workload, opts: Options) -> an
 
     #[cfg(target_os = "linux")]
     {
-        if let Some(skip_after) = &opts.flamegraph_options.skip_after {
-            collapse_options.skip_after = Some(skip_after.into())
-        }
+        collapse_options.skip_after = opts.flamegraph_options.skip_after.clone();
```
djc, line 296: Why do we need to clone here?

djc (commented):
Hey, thanks for fixing this up!

src/lib.rs
martinxyz: Because a few lines below, `into_inferno()` fully consumes `flamegraph_options`, despite not being interested in the `.skip_after` field.

I think it was doing a copy before already. We could probably drain/reassign the vector to make it clear it isn't going to be used again, but it also seems wrong to modify the command-line options.

djc:
Thanks!

PR #200, opened 2022-04-10
Fix running with --root on Linux.

### LINKED ISSUE #99: Getting a panic when using the `--root` option
Hello!

I've been trying to run `cargo flamegraph`.

Unprivileged, it gave me an error complaining about `kernel.perf_event_paranoid = -1`, which I promptly solved by creating a `/etc/sysctl.d/10-cargo-flamegraph.conf` file with the aforementioned line.

After that, it began giving me the following warning:

```
WARNING: Kernel address maps (/proc/{kallsyms,modules}) are restricted,
check /proc/sys/kernel/kptr_restrict and /proc/sys/kernel/perf_event_paranoid.

Samples in kernel functions may not be resolved if a suitable vmlinux
file is not found in the buildid cache or in the vmlinux path.

Samples in kernel modules won't be resolved at all.

If some relocation was applied (e.g. kexec) symbols may be misresolved
even with a suitable vmlinux or kallsyms file.

Couldn't record kernel reference relocation symbol
Symbol resolution may be skewed if relocation was used (e.g. kexec).
Check /proc/kallsyms permission or run as root.
```

A flamegraph was generated but there was mostly just `[unknown]` sprinkled all over the graph.

At this point, I ran into the `--root` option, which seemed like a good option to bypass all these problems. But when I used it, another entirely different problem happened:

```
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.180 MB perf.data (20 samples) ]
thread 'main' panicked at 'unable to generate a flamegraph from the collapsed stack data: Io(Custom { kind: InvalidData, error: "No stack counts found" })', /home/valmirpretto/.cargo/registry/src/github.com-1ecc6299db9ec823/flamegraph-0.3.1/src/lib.rs:273:6
```

Is there something I'm missing?

I'm trying to run `cargo flamegraph` on an Ubuntu machine, if that's relevant.

mstange:
Ensure that the perf.data file generated by `perf record` is readable by`perf script`.
Fixes #99.

This just copies the very similar code which handles this in the dtrace case.

This issue was identified in https://github.com/flamegraph-rs/flamegraph/issues/99#issuecomment-827437480 by @jscarrott:
> It's caused by perf.data ending up being owned by root with the current user not having any permissions to read the file.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,8 +93,22 @@ mod arch {
     pub fn output(
         perf_output: Option<String>,
         script_no_inline: bool,
-        _sudo: bool,
+        sudo: bool,
     ) -> anyhow::Result<Vec<u8>> {
+        if sudo {
+            // We executed `perf record` with sudo, and will be executing `perf script` *without* sudo.
+            // Ensure the perf.data file is readable by the current user, so that `perf script` can
+            // read it.
+            if let Ok(user) = env::var("USER") {
+                Command::new("sudo")
+                    .args(["chown", user.as_str(), "perf.data"])
+                    .spawn()
+                    .expect(arch::SPAWN_ERROR)
+                    .wait()
+                    .expect(arch::WAIT_ERROR);
+            }
+        }
+
         let perf = env::var("PERF").unwrap_or_else(|_| "perf".to_string());
         let mut command = Command::new(perf);
 


djc:
Are people at Mozilla using flamegraph? Would you be interested in helping maintain this repo?

djc (approved):
LGTM, thanks!

djc:
> I just personally care about the profiling situation in the general Rust ecosystem, so this project is interesting to me because it seems to have a large existing audience. I like to work on profilers in my spare time.

Well, your contributions are very welcome! I'm sort of a passive maintainer who sometimes has a use for flamegraph, but I use macOS as my primary development environment and in my experience flamegraph doesn't work all that well with the macOS dtrace.

> Not at this point, thank you. I just became co-maintainer of another repo and have yet to actually do any maintaining on it, so I don't want to put too much on my plate.

Fair enough! Let me know if that ever changes.

PR #202, opened 2022-05-07
Add cargo-flamegraph windows support

nico-abram:
Fixes https://github.com/flamegraph-rs/flamegraph/issues/3

I recently wrote [a library crate](https://crates.io/crates/blondie) to do callstack sampling on windows. It can output the same format that using dtrace as a sampling profiler generates. It uses the [ETW](https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/event-tracing-for-windows--etw-) windows API (More details in the [readme](https://github.com/nico-abram/blondie/))

It still requires running as administrator, like DTrace, but does not require being part of the "Windows Insider Program" or installing a separate program like DTrace and should also work on windows 7 and 8.

This patch
 - Only works for the "Command" workloads, which I think is all that's needed for the cargo-flamegraph binary
 - Uses a hacky workaround to change as little code as possible and deal with the fact that the sampling library wants to spawn the Command for us.
 - Leaves dtrace support intact, and only attempts to run with the new provider if dtrace is not found. To do this, it runs "$(dtrace_command) --help" to test if dtrace is available.

I'm opening this as a draft to see what you think. If you'd be willing to eventually merge a similar patch to this before working on it further, I have some questions:
 - Would it be a requirement to add support for Pid workloads before merging?
 - Would it be a requirement to correctly support a custom frequency?
 - Would it be OK to move the Command spawn and wait calls into `arch::initial_command` and remove the Command from the return tuple? That way I could get rid of the "cmd /C exit 0" workaround.
 - How important would Windows 7/8 support be? I plan to test how well this patch works on them, but I'm asking because some ETW APIs I might want to use for improvements are only available since windows 10.

Here's an example of what the generated flamegraph looks like, running on flamegraph itself without arguments, which just printed the help message:
imagen

The library also samples kernel callstacks, and can do symbol resolution for them, but since they add a lot of noise and can be confusing for most users it has a boolean parameter to hide them, and this patch hardcodes that to disabled.


djc:
Sorry, marked this as done by accident. This sounds very interesting, will review in more detail soon!

djc:
In general, as I understand it the dtrace backend on Windows is more or less non-functional out of the box. Therefore any functionality added that enables use cases on Windows is strictly additive and would be welcome. As such, there's no need to be complete in terms of PID vs running a process or OS version support in this first PR — we can iterate to extend support.

It does sound great if you can refactor the code to avoid the unnecessary Command call. Ideally the refactoring would be in a separate commit.

djc:
(Ideally the clippy commit gets squashed into the commit that introduces the warning.)

nico-abram:
Squashed the lint fix commit. I left it in earlier since I figured it'd make review easier.
Added a second commit to refactor initial_command.

I tried this on a 32bit windows 7 VM and after [a couple changes](https://github.com/nico-abram/blondie/commit/094a3daab268276e431ace42c386fe2a9bc91e6d) to the sampling library it sampled the callstacks just fine, but then panicked in inferno because of an underflow somewhere that is using usize.

src/lib.rs
```diff
@@ -87,7 +99,17 @@ mod arch {
             Workload::ReadPerf(_) => (),
         }
 
-        (command, perf_output)
+        if verbose {
```
djc, line 102: How complex would it be to deduplicate the two copies of this code?

nico-abram: I could factor it out into a function as

```rs
                        if verbose {
                            print_command(&command);
                        }
```

Or maybe pass in a closure to `arch::initial_command` such as `if opts.verbose { |com| println!("command: {:?}", com); } else { |_| {} }`

nico-abram: Factored it out to a function for now

djc:
Thanks! I'm going to follow up with some further refactoring to try to reduce duplication some more.

Eh2406:
When this is ready for use, can we update the readme?

djc:
@nico-abram would you be able to do that?

LoganDark:
> When this is ready for use, can we update the readme?

image

For me it just hangs here, I don't think it's ready yet.

nico-abram:
@LoganDark  This PR added an alternativa way of collecting samples other than dtrace. Try removing dtrace from your path, I made It prioritize dtrace to avoid breaking existing users, but maybe that's a bad idea since dtrace has never worked well on windows

LoganDark:
> @LoganDark This PR added an alternativa way of collecting samples other than dtrace. Try removing dtrace from your path, I made It prioritize dtrace to avoid breaking existing users, but maybe that's a bad idea since dtrace has never worked well on windows

```
Error: unable to generate a flamegraph from the collapsed stack data

Caused by:
    0: I/O error: No stack counts found
    1: No stack counts found
```

Great it's the same error I get trying to use the CLion profiler in WSL2. I give up, I fucking hate Windows.

nico-abram:
> @nico-abram would you be able to do that?

I can, but I wanna make some changes suggested by people from firefox-profiler to symbol resolution first (In https://github.com/nico-abram/blondie/issues/2, basically better performance for big binaries) in the library and I haven't had the time yet.

nico-abram:
> > @LoganDark This PR added an alternativa way of collecting samples other than dtrace. Try removing dtrace from your path, I made It prioritize dtrace to avoid breaking existing users, but maybe that's a bad idea since dtrace has never worked well on windows
> 
> ```
> writing flamegraph to "flamegraph.svg"
> Error: unable to generate a flamegraph from the collapsed stack data
> 
> Caused by:
>     0: I/O error: No stack counts found
>     1: No stack counts found
> ```

Thanks for testing this change!
Do you mind if I ask you some questions to try to find where the problem is coming from? I'd love to fix this!

1) Did you compile cargo flamegraph yourself? I don't think this change has shipped in a release yet
2) Does the program you're debugging have debug info?
3) Is the program you're debugging open source/available for testing?
4) If you want to try debugging the problem, you could add something like `dbg!(&collapsed)` [here](https://github.com/flamegraph-rs/flamegraph/blob/d08ddb23662e426e15c234c31464f7f8b9e778dd/src/lib.rs#L373) and `dbg!(&output);` [here](https://github.com/flamegraph-rs/flamegraph/blob/d08ddb23662e426e15c234c31464f7f8b9e778dd/src/lib.rs#L355). That might generate a lot of output to stderr, so you might want to save that to a file (Something like `cargo flamegraph 2>&1 > out.txt`) and share it here. 

> Great it's the same error I get trying to use the CLion profiler in WSL2. I give up, I fucking hate Windows.

*Please* try to remain civil. Maybe the WSL2 is because hardware performance counters don't work in WSL2, although I heard that was fixed [in november](https://github.com/microsoft/WSL/issues/4678) but I haven't tried it.

You might have better luck with WPR/WPA(Windows Performance Recorder/Analyzer) but they can be a bit hard to use. The Visual Studio profiler should be easier to use, but gives a bit less information (Open VS, file>Open>Project, open the .exe binary, configure the arguments and cwd in the debugging options ( Debug>Debug Properties), then Debug>Perf profiler>cpu usage>start. Make sure you have debug info i.e `debug=true` in your cargo.toml).

LoganDark:
> 1\. Did you compile cargo flamegraph yourself? I don't think this change has shipped in a release yet

Yes, I used `cargo install --git https://github.com/flamegraph-rs/flamegraph`

> 2\. Does the program you're debugging have debug info?

Yes, I am using release mode with `CARGO_PROFILE_RELEASE_DEBUG=true`, but dev mode has the same issue

> 3\. Is the program you're debugging open source/available for testing?

It happens for any program, so I dunno if that's the problem, but see https://github.com/j-tai/getargs for the library I'm trying to profile (the `no_alloc` example specifically)

> 4\. If you want to try debugging the problem, you could add something like `dbg!(&collapsed)` [here](https://github.com/flamegraph-rs/flamegraph/blob/d08ddb23662e426e15c234c31464f7f8b9e778dd/src/lib.rs#L373) and `dbg!(&output);` [here](https://github.com/flamegraph-rs/flamegraph/blob/d08ddb23662e426e15c234c31464f7f8b9e778dd/src/lib.rs#L355). That might generate a lot of output to stderr, so you might want to save that to a file (Something like `cargo flamegraph 2>&1 > out.txt`) and share it here.

```
[src\lib.rs:355] &output = []
[src\lib.rs:373] &collapsed = []
```

> Maybe the WSL2 thing is because hardware performance counters don't work in WSL2, although I heard that was fixed [in november](https://github.com/microsoft/WSL/issues/4678) but I haven't tried it.

I compiled `perf` from source and installed it as Microsoft/IntelliJ-Rust recommends (but this is off-topic)

> You might have better luck with WPR/WPA(Windows Performance Recorder/Analyzer) but they can be a bit hard to use. The Visual Studio profiler should be easier to use, but gives a bit less information (Open VS, file>Open>Project, open the .exe binary, configure the arguments and cwd in the debugging options ( Debug>Debug Properties), then Debug>Perf profiler>cpu usage>start. Make sure you have debug info i.e `debug=true` in your cargo.toml).

I'll look into it, thanks.

nico-abram:
`cargo flamegraph  --example no_alloc` worked for me, although I only got 40 samples
imagen

Did you run cargo-flamegraph as administrator? What windows version are you on?

Similar to what you saw with CLion according to your gist. I tried VS and I got more or less the same, only 36 samples: 
imagen

WPA/WPR show callstacks for me, although with only 41 samples:
imagen


I recorded with these settings in WPR:
imagen
Drag "Computation" into the analysis tab:
imagen

Then filter to no_alloc:
imagen

Then load symbols:
imagen

You can also use this flame view in WPA:
imagen

Which looks like this:
imagen

I get the feeling most of the time, if any, is spent in I/O and that's why CPU sampling is not turning up much.

LoganDark:
> Did you run cargo-flamegraph as administrator? What windows version are you on?

I did run it as Administrator. I'm on Windows 10 Home 2004 build 19041.928 (64-bit Intel).

> I get the feeling most of the time, if any, is spent in I/O and that's why CPU sampling is not turning up much.

Most of my testing was done with a fork of the program that:

- Does not output to stderr (i.e. all `eprintln!` removed)
- Loops 10,000 times

In that case, there is absolutely no reason that the majority of the time should be spent in the kernel, the profiler results are not consistent with reality.

Given what you've seen in the Gist, is there any way to make `cargo flamegraph` work here, in WSL2 or on native Windows? Also, if you think there's a more appropriate place for my question, I'm all ears. I've already tried at least 3 times (on different days) in the Rust Discord server, on some other programming/sysadmin-related servers, and on various Microsoft & WSL issue trackers. No dice, no dice, no dice.

nico-abram:
I think it's likely to be because of bugs like these https://developercommunity.visualstudio.com/t/No-Data-in-CPU-Usage-Tool-Windows-Updat/1384519 / https://developercommunity.visualstudio.com/t/No-Data-in-CPU-Usage-Tool-Windows-Kerne/1573253 and running and old windows 10 version where they are still present

LoganDark:
image

That was the exact issue, updating my Windows Defender threat definitions (even though it wasn't actually enabled??) got the profiler working!! Yay!

SuhasHebbar:
> image
> 
> That was the exact issue, updating my Windows Defender threat definitions (even though it wasn't actually enabled??) got the profiler working!! Yay!

@LoganDark,
I also seem to be facing a similar issue as you on Windows. 
Could you elaborate on how fixing "Windows Defender threat definitions" fixed the issue for you?

Blondie doesn't seem to be detecting any stack walk events at all for me.

PR #206, opened 2022-06-03
Fix examples for arbitrary process

Niederb:
The examples for arbitrary processes seems wrong. At least I had to change them to make them work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,13 +68,13 @@ flamegraph --completions bash > $XDG_CONFIG_HOME/bash_completion # or /etc/bash_
 
 ```
 # if you'd like to profile an arbitrary executable:
-flamegraph [-o my_flamegraph.svg] /path/to/my/binary --my-arg 5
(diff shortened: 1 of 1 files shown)


djc:
Hmm, that's a little surprising, I guess this might be a regression from migrating to clap? Would you be able to check if this can fixed with our declarative clap setup instead?

Niederb:
I just had a quick look at the clap documentation and this is documented behavior. `trailing_arguments` is marked as `last=true`:
```
    #[clap(last = true)]
    trailing_arguments: Vec<String>
```
And from the clap-doc (https://docs.rs/clap/3.1.18/clap/struct.Arg.html#method.last)

> This arg is the last, or final, positional argument (i.e. has the highest index) and is only able to be accessed via the -- syntax 

I don't know if there would be another way to get it working without `--`.

djc:
So could we leave off the `last = true`?

lukehsiao:
https://github.com/flamegraph-rs/flamegraph/pull/218/files

It looks like one of these changes already made it into main, so only the latter is relevant.

PR #210, opened 2022-07-19
Add --test example to readme

kreibaum:
I was struggling with figuring out how to run an integration test with flamegraph for a bit. I think adding an example for it to the readme makes it clearer that --unit-test and --test are separate and I suppose that would have helped me.


djc:
Makes sense, thanks!

PR #215, opened 2022-07-25
Add documentation about `--pid` flag

### LINKED ISSUE #22: extend flamegraph to take an optional pid and timespan
Adding a timeout to `flamegraph` would be nice for long running commands.

Example of a command printing output once per second forever:
`# flamegraph -o example.svg vfsstat 1`

In this scenario being able to pass a 30s timeout value that results in `tick-30s { exit(0) }'` would be great.
`# flamegraph -t 30s -o example.svg vfsstat 1`

Additionally being able to run `flamegraph` against an already running process by specifying a pid and timeout would be nice.
`flamegraph -p $(pgrep my-server) -t 30s -o server.svg`

ckcr4lyf:
Quite a useful feature, even wanted by some, turns out exists all along!

Just not documented, so adding it here.

~Fixes #22  (kinda?)~ (actually no, that one also talks about a timeout value)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,9 @@ flamegraph --completions bash > $XDG_CONFIG_HOME/bash_completion # or /etc/bash_
 # if you'd like to profile an arbitrary executable:
 flamegraph [-o my_flamegraph.svg] /path/to/my/binary --my-arg 5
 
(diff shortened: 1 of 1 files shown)


djc:
Thanks!

Issue #214, opened 2022-07-25
Add option to post-process the folded stacks with a command.


hal8174:
# Problem

If you want to be able to post process the folded stacks, you can't use inferno directly. In my case, I wanted to profile a recursive function. To do this, I want to use something like this [stackcollapse-recusive.pl](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-recursive.pl) from the original flamegraph.

# My proposed solution

Add option to post-process the folded stack-frames.

```
cargo flamegraph --post-process command
```
`command` is executed receiving the folded stacks via stdin and outputting the post-processed samples to stdout.

In my example, I would have a command that collapses recursive stacks and could just call `cargo flamegraph --post-process collapse-recusive` to create the desired outcome.

----

Is this the correct way to implement post-processing folded stacks, or should this done somewhere else.

I would implement this feature, but wanted to check if this is something, that would be added.

djc:
I would recommend opening an issue in the inferno repository to see if they would be interested in having the option to collapse recursion inside inferno. If inferno exposes logic for that, we can then wire that up in the flamegraph interface.

cc @jonhoo

jonhoo:
I'm not entirely sure I understand the ask from inferno. With the command-line interface you can already do this by just filtering the output of `inferno-collapse-*` through `collapse-recursive` before passing it to `inferno-flamegraph`. With the library interface you'd do such a "mid-processing" step after calling the appropriate `collapse::` implementation but before passing it to `flamegraph::from_*`. I'm not entirely sure how `flamegraph-rs` calls into inferno, but in either case it should be possible to modify the output of the collapse step before charting. I don't think `inferno` will itself get a built-in `collapse-recursive` — it's nice for those kinds of tweaks to be expressible as plug-and-play line-based text filtering tools.

djc:
Well, I think you covered all the bases! In that case, I'm open to having something like the proposed `--post-process` option in flamegraph.

hal8174:
I started implementing this feature, but ran into two problems:

### Argument Parsing
There are basically two options:

1. `--post-process "command --option1 --option2"`
2. `--post-process command --option1 --option2`

The advantage of the first option is that it is very clear which options are for which command. The disadvantage is that internally it would have to be split to be passed to the `Command`. This might require a crate like [shlex](https://crates.io/crates/shlex). I couldn't find if this is possible with clap.
[...]

PR #216, opened 2022-07-27
Add --post-process feature

hal8174:
Implementation of the feature suggested in #214.

I decided to provide the command as a string and split it into arguments using [shlex](https://crates.io/crates/shlex). A problem with this is, that shlex splits the string, mimicking a POSIX shell. I don't know enough about windows to know if that could be an issue.

The difficulty with deadlocks is solved by using two threads.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,6 +206,7 @@ dependencies = [
  "clap_complete",
  "inferno",
  "opener",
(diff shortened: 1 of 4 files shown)


djc:
This seems reasonable to me! If someone runs into problems on Windows they'll report back, I guess. 👍

PR #218, opened 2022-08-11
fix(readme): add missing -- to call on binary

IceTDrinker:
Tried to use the tool and the -- seems to be missing from the command line given as an example


djc:
Thanks!

Issue #220, opened 2022-09-01
Instructions should state Debian stable is not supported


richardg867:
Despite README.md containing instructions for installing dependencies on Debian, flamegraph itself cannot be installed on Debian stable (currently bullseye) due to the outdated Rust version it packages. README.md should state the minimum required Debian version, or at least provide instructions on installing a newer version of Rust (maybe a link to rustup).

```
    Updating crates.io index
  Downloaded flamegraph v0.6.2
  Downloaded 1 crate (640.8 KB) in 0.54s
  Installing flamegraph v0.6.2
  Downloaded cfg-if v1.0.0
  Downloaded autocfg v1.1.0
  Downloaded quote v1.0.21
  Downloaded version_check v0.9.4
  Downloaded opener v0.5.0
  Downloaded regex-automata v0.1.10
  Downloaded bstr v0.2.17
  Downloaded nodrop v0.1.14
  Downloaded anyhow v1.0.63
  Downloaded arrayvec v0.4.12
  Downloaded camino v1.1.1
  Downloaded bitflags v1.3.2
  Downloaded quick-xml v0.23.0
  Downloaded clap v3.2.19
error: failed to compile `flamegraph v0.6.2`, intermediate artifacts can be found at `/tmp/cargo-installFinW6v`

Caused by:
  failed to parse manifest at `/root/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.2.19/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.
```

```
$ rustc --version
rustc 1.48.0

$ sudo apt install rustc
[...]
rustc is already the newest version (1.48.0+dfsg1-2).
```

richardg867:
#221 submitted.

PR #221, opened 2022-09-02
readme: Add note about the Rust version on Debian bullseye

richardg867:
Requested by #220. Debian bullseye (the current stable) packages rustc 1.48.0 which does not support Rust 2021.


djc (approved):
Thanks!

PR #224, opened 2022-09-09
fix: correct guidance for debuginfo and unit tests

dkales:
Currently, when calling `cargo flamegraph --unit-test -- path::to::test` one gets a warning to enable debug symbols for the bench profile:

```
WARNING: profiling without debuginfo. Enable symbol information by adding the following lines to Cargo.toml:

[profile.bench]
debug = true

Or set this environment variable:

CARGO_PROFILE_BENCH_DEBUG=true
```
However, unit tests are actually run using `cargo test --release` internally, so the above message should reference `profile.release`.

This PR extends the check on what to print for unit tests.

Sidenote: The `or_else` in check below is to adhere to clippy, since the call is too long for a simple or, happy if someone has a more concise alternative.


djc:
Thanks, this seems reasonable!

PR #227, opened 2022-09-15
Improve 'perf script' failure error

adambratschikaye:
Report an error when running `perf script` on the `perf.data` file fails. Currently when `perf script` fails it just passes empty data on to `inferno` which results in an unclear error about missing stack data.

Instead, this change checks the exit status of `perf script` and immediately exits if it failed.

This fixes https://github.com/flamegraph-rs/flamegraph/issues/226.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -124,10 +124,15 @@ mod arch {
             command.arg(perf_output);
         }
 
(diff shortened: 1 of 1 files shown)


djc (approved):
Looks great, thanks!

PR #233, opened 2022-10-06
Use default_run from Cargo.toml as default bin target

### LINKED ISSUE #232: Allow omitting default binary target
The readme indicates that `cargo flamegraph` is equivalent to `cargo run --release`. However, in my project with multiple binary targets it gives the following error:
```
Error: several possible targets found: [BinaryTarget { package: "astar-pairwise-aligner", target: "astar-pairwise-aligner", kind: ["bin"] }, BinaryTarget { package: "astar-pairwise-aligner", target: "proof", kind: ["bin"] }, BinaryTarget { package: "astar-pairwise-aligner", target: "generate_dataset", kind: ["bin"] }], please pass an explicit target.
```

It would be convenient to run the default binary (ie the name of the crate / the one set as `default-run` in `cargo.toml`).

Additionally it would be nice to allow the `--release` flag as a no-op, so that swapping `run` for `flamegraph` in `cargo run -r -- <flags>` _just works_.

RagnarGrootKoerkamp:
Fix #232 

This hides the info message when there is a single package with a `default_run` set. The code to do that came out slightly verbose though.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -310,18 +310,34 @@ fn find_unique_target(
         });
     }
 
+    let mut num_packages = 0;
+    let mut is_default = false;
+
     let mut targets: Vec<_> = packages
         .flat_map(|p| {
-            let Package { targets, name, .. } = p;
+            let Package {
+                targets,
+                name,
+                default_run,
+                ..
+            } = p;
+            num_packages += 1;
+            if default_run.is_some() {
+                is_default = true;
+            }
             targets.into_iter().filter_map(move |t| {
-                t.kind
-                    .iter()
-                    .any(|s| kind.contains(&s.as_str()))
-                    .then(|| BinaryTarget {
-                        package: name.clone(),
-                        target: t.name,
-                        kind: t.kind,
-                    })
+                // Keep only targets that are of the right kind.
+                let ok_kind = t.kind.iter().any(|s| kind.contains(&s.as_str()));
+                // When `default_run` is set, keep only the target with that name.
+                let default_filter = match &default_run {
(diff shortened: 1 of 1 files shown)


src/bin/cargo-flamegraph.rs
```diff
@@ -310,29 +310,45 @@ fn find_unique_target(
...
 
     match targets.as_slice() {
         [_] => {
             let target = targets.remove(0);
-            eprintln!(
-                "automatically selected {} as it is the only valid target",
-                target
-            );
+            // If the selected target is the default_run of the only package, do not print a message.
+            if !(num_packages == 1 && is_default) {
```
djc, line 346: I'd prefer the `!= || !` formulation over `!(== && )`.

src/bin/cargo-flamegraph.rs
```diff
@@ -310,29 +310,45 @@ fn find_unique_target(
...
-                    .any(|s| kind.contains(&s.as_str()))
-                    .then(|| BinaryTarget {
-                        package: name.clone(),
-                        target: t.name,
-                        kind: t.kind,
-                    })
+                let ok_kind = t.kind.iter().any(|s| kind.contains(&s.as_str()));
+                let default_filter = default_run
+                    .as_ref()
+                    .map_or(true, |default_name| &t.name == default_name);
```
djc, line 332: I'd like this rewritten as a `match` expression. Also probably good to add some comments inside the `filter_map()` closure here about what it is we're trying to achieve.

djc (commented):
Thanks, some small suggestions.

RagnarGrootKoerkamp:
done

djc (approved):

PR #239, opened 2022-10-24
Moved install command from Pop!_OS to common section.

ajgrah2000:
Added raspberry pi perf apt command.
Removed legacy note about cargo-flamegraph:
  The command doesn't seem to work,
  the note will still be in git history if need be.


ajgrah2000:
Main intent was to put the 'cargo install flamegraph' in what subjectively appears to be a clearer location (happy to drop the other changes if they're note appropriate).

djc (approved):
Thanks!

PR #241, opened 2022-11-02
Add support for cargo --profile

### LINKED ISSUE #167: Support custom cargo profiles
Now that https://github.com/rust-lang/cargo/issues/6988 is stabilized, it might be nice if it were possible to specify any profile rather than just --dev or --release. This was we can have an optimised debug release which is probably optimal for flamegraphing:

```
[profile.release-debug]
lto = "fat"
codegen-units = 1
debug = true
```

madadam:
Closes #167

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -12,6 +12,10 @@ struct Opt {
     #[clap(long)]
     dev: bool,
 
+    /// Build with the specified profile
+    #[clap(long)]
+    profile: Option<String>,
+
     /// package with the binary to run
     #[clap(short, long)]
     package: Option<String>,
@@ -84,8 +88,10 @@ fn build(opt: &Opt, kind: impl IntoIterator<Item = String>) -> anyhow::Result<Ve
         cmd.arg("build");
     }
 
-    // do not use `--release` when we are building for `bench`
-    if !opt.dev && opt.bench.is_none() {
+    if let Some(profile) = &opt.profile {
+        cmd.arg("--profile").arg(profile);
+    } else if !opt.dev && opt.bench.is_none() {
+        // do not use `--release` when we are building for `bench`
         cmd.arg("--release");
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -501,7 +501,7 @@ pub struct FlamegraphOptions {
     pub notes: Option<String>,
 
     /// Omit functions smaller than <FLOAT> pixels
-    #[clap(long, default_value = "0.01f32", value_name = "FLOAT")]
+    #[clap(long, default_value = "0.01", value_name = "FLOAT")]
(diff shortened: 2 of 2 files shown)


madadam:
Question: should we also deprecate / remove the `--dev` flag?

src/bin/cargo-flamegraph.rs
```diff
@@ -84,9 +88,13 @@ fn build(opt: &Opt, kind: impl IntoIterator<Item = String>) -> anyhow::Result<Ve
         cmd.arg("build");
     }
 
-    // do not use `--release` when we are building for `bench`
-    if !opt.dev && opt.bench.is_none() {
-        cmd.arg("--release");
+    if let Some(profile) = &opt.profile {
+        cmd.arg("--profile").arg(profile);
+    } else {
```
djc, line 93: Would prefer to use `else if` directly here.

src/lib.rs
```diff
@@ -501,7 +501,7 @@ pub struct FlamegraphOptions {
     pub notes: Option<String>,
 
     /// Omit functions smaller than <FLOAT> pixels
-    #[clap(long, default_value = "0.01f32", value_name = "FLOAT")]
+    #[clap(long, default_value = "0.01", value_name = "FLOAT")]
```
djc, line 504: This seems unrelated?

djc:
> Question: should we also deprecate / remove the `--dev` flag?

No, let's keep that for now.

src/lib.rs
madadam: Yes, it was throwing runtime error for me, probably because "0.01f32" doesn't parse to `f64`? But you are right, it's unrelated. Should I put it into a separate commit?

djc: Yes, splitting it out in a separate commit would be fine, thanks!

src/bin/cargo-flamegraph.rs
```diff
@@ -84,8 +88,10 @@ fn build(opt: &Opt, kind: impl IntoIterator<Item = String>) -> anyhow::Result<Ve
         cmd.arg("build");
     }
 
+    if let Some(profile) = &opt.profile {
+        cmd.arg("--profile").arg(profile);
     // do not use `--release` when we are building for `bench`
```
djc, line 93: Not sure if this will pass `cargo fmt`, the comment placement seems a bit weird here.

madadam: This is actually why I originally didn't collapse the `else if`.

madadam: btw this is already has `cargo fmt` applied

djc: Okay, let's just move the comment inside the `else if`? That seems like a cleaner way to clarify.

djc (approved):

djc:
Thanks!

PR #242, opened 2022-11-17
Correctly identify named unit test targets

### LINKED ISSUE #211: --unit-test flag doesn't allow profiling unit tests in a lib-only crate in a workspace
To repro, create a new cargo workspace with two crates containing only lib targets (you can use them exactly as created by `cargo new`).

If you just pass the `--unit-test` flag, it fails because it doesn't know which target to build
```
$ cargo flamegraph --unit-test -- tests::it_works
Error: several possible targets found: [BinaryTarget { package: "subproject_one", target: "subproject_one", kind: ["lib"] }, BinaryTarget { package: "subproject_two", target: "subproject_two", kind: ["lib"] }], please pass an explicit target.
```

However, if I _do_ provide a target argument, it errors out because it is expecting a bin target, and only finds a lib target with the specified name.
```
cargo flamegraph --unit-test subproject_one -- tests::it_works
error: no bin target named `subproject_one`
Error: cargo build failed
```

Is this expected behavior, or a bug? I would expect that with the `--unit-test` flag specified, it shouldn't care whether the target is bin or lib, since it's building a test binary to profile

ramosbugs:
The `--unit-test` flag (with no argument) only works in crates with a single target. In crates with multiple targets, the flag results in a `several possible targets found` error. If a specific target name is passed as an argument to `--unit-test`, the tool always passes the `--bin` flag to Cargo, which fails for `lib` targets.

This PR changes the behavior when an argument is passed to `--unit-test` to resolve the name to the corresponding target in order to determine whether to pass the `--lib` or `--bin` flag to Cargo.

Fixes #211.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -344,9 +344,10 @@ fn find_unique_target(
                     None => true,
                     Some(default_name) => &t.name == default_name,
                 };
-                let name_filter = target_name
-                    .map(|target_name| t.name == target_name)
-                    .unwrap_or(true);
+                let name_filter = match target_name {
+                    None => true,
+                    Some(target_name) => t.name == target_name
+                };
                 (ok_kind && default_filter && name_filter).then(|| BinaryTarget {
                     package: name.clone(),
                     target: t.name,
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -346,7 +346,7 @@ fn find_unique_target(
                 };
                 let name_filter = match target_name {
                     None => true,
-                    Some(target_name) => t.name == target_name
+                    Some(target_name) => t.name == target_name,
                 };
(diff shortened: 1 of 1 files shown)


src/bin/cargo-flamegraph.rs
```diff
@@ -343,7 +344,10 @@ fn find_unique_target(
                     None => true,
                     Some(default_name) => &t.name == default_name,
                 };
-                (ok_kind && default_filter).then(|| BinaryTarget {
+                let name_filter = target_name
+                    .map(|target_name| t.name == target_name)
+                    .unwrap_or(true);
```
djc, line 349: I find the spelling like `default_filter` (with an explicit `match`) a good deal easier to follow than the `map().unwrap_or()` combinators. Would you mind rewriting it that way?

djc (commented):
Thanks for fixing this! I have a stylistic concern but otherwise this LGTM.

djc:
Sorry -- add the comma, please?

djc:
Thanks!

PR #250, opened 2023-01-18
Added a command line option to ignore perf's return status

lorlouis:
Adds a `--ignore-status` flag which ignores perf's exit code.

#### why?

I was trying to use flamegraph on a failing test and flamegraph would fail with "failed to sample program" even though perf was able to sample the program because perf would return the same exit code as the program it was running (a failing test), this confused flamegraph and it though that perf was not able to sample the program.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,49 +124,35 @@ cargo flamegraph --test test_name
 `flamegraph` is quite simple. `cargo-flamegraph` is more sophisticated:
 
 ```
-USAGE:
-    cargo-flamegraph flamegraph [FLAGS] [OPTIONS] [trailing-arguments]...
+Usage: flamegraph [OPTIONS] [-- <TRAILING_ARGUMENTS>...]
 
-FLAGS:
-        --deterministic          Colors are selected such that the color of a function does not change between runs
-        --dev                    Build with the dev profile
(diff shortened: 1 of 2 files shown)


djc (approved):
Makes sense, thanks!

README.md
```diff
@@ -124,49 +124,35 @@ cargo flamegraph --test test_name
...
-        --skip-after <FUNCTION>            Cut off stack frames below <FUNCTION>; may be repeated [linux only]
-        --test <test>                      Test binary to run (currently profiles the test harness and all tests in the
-                                           binary)
-        --unit-test <unit-test>            Crate target to unit test, <unit-test> may be omitted if crate only has one
-                                           target (currently profiles the test harness and all tests in the binary; test
-                                           selection can be passed as trailing arguments after `--` as separator)
-
-ARGS:
-    <trailing-arguments>...    Trailing arguments are passed to the binary being profiled
+Usage: flamegraph [OPTIONS] [-- <TRAILING_ARGUMENTS>...]
```
djc, line 127: This updates from using `cargo-flamegraph` to `flamegraph`, I think we should keep the former?

lorlouis: Nice catch I've regenerated the usage message using `cargo flamegraph --help`

djc:
Thanks!

PR #251, opened 2023-02-10
Update blondie to v0.4

nico-abram:
Fixes https://github.com/flamegraph-rs/flamegraph/issues/207

Blondie now uses the [pdb-addr2line](https://crates.io/crates/pdb-addr2line) and [symsrv](https://crates.io/crates/symsrv) crates instead of microsoft's dbghelp.dll for symbol resolution. This is faster, and resolved a couple bugs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ shlex = "1.1.0"
 signal-hook = "0.3.10"
 
 [target.'cfg(windows)'.dependencies]
-blondie = "0.3.0"
+blondie = "0.4.0"
 
 [profile.release.build-override]
 opt-level = 0
--- a/deny.toml
+++ b/deny.toml
@@ -2,3 +2,15 @@
 allow-osi-fsf-free = "either"
 copyleft = "warn"
 allow = ["CDDL-1.0"]
+
+[[licenses.clarify]]
+name = "ring"
+# SPDX considers OpenSSL to encompass both the OpenSSL and SSLeay licenses
+# https://spdx.org/licenses/OpenSSL.html
+# ISC - Both BoringSSL and ring use this for their new files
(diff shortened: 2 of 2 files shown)


djc:
You'll need to add a blurb to the `deny.toml` for the *ring* license:

```toml
[[licenses.clarify]]
name = "ring"
expression = "ISC AND MIT AND OpenSSL"
license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
```

What's the reason for the extensive list of licenses that you introduced in https://github.com/nico-abram/blondie/commit/965b4130c7120392f0e17fcc0d9abd12e1b3d898? In particular, the MPL-2.0 is considered copyleft, which can be tricky.

djc:
(I pushed a change to update the `Cargo.lock` file with recent dependencies, hopefully this helps with some of the cargo-deny issues.)

nico-abram:
> What's the reason for the extensive list of licenses that you introduced in [nico-abram/blondie@965b413](https://github.com/nico-abram/blondie/commit/965b4130c7120392f0e17fcc0d9abd12e1b3d898)? In particular, the MPL-2.0 is considered copyleft, which can be tricky.

My understanding was offering multiple licenses in that manner allowed the consumer to pick the one that best suits their needs, so aiming to make it maximumly available I added some of the most commonly accepted licenses. If that complicates things I can remove MPL

djc:
Thanks!

> My understanding was offering multiple licenses in that manner allowed the consumer to pick the one that best suits their needs, so aiming to make it maximumly available I added some of the most commonly accepted licenses. If that complicates things I can remove MPL

I would recommend sticking with the default of just `MIT OR Apache-2.0`. Given that MIT already allows most uses (with attribution), this combo should be fine for pretty much everyone (there are good reasons this is the default in the Rust ecosystem).

PR #255, opened 2023-03-07
[WIP] feat: add --iterations flags to specify how many times to run target binary

bartlomieju:
Hi there! Thanks for the great project, we use it extensively at Deno. 

While we are working on improving our startup time, we found out that running the target binary just once is not enough to collect reliable data to see where we're spending time. So I added `--iterations` flag that allows to run target binary multiple times and create a single flamegraph out of all the runs.

This is a quick a dirty implementation, but it allows us to collect more detailed data. If there's an interest in adding this flag and upstreaming the patch I'll be happy to clean it up.


Cargo.lock
```diff
@@ -2,6 +2,12 @@
 # It is not intended for manual editing.
 version = 3
 
+[[package]]
```
djc, line 5: These changes don't belong in this PR, please revert.

src/lib.rs
```diff
@@ -316,7 +316,7 @@ fn print_command(cmd: &Command, verbose: bool) {
     }
 }
 
-pub fn generate_flamegraph_for_workload(workload: Workload, opts: Options) -> anyhow::Result<()> {
+pub fn generate_flamegraph_for_workload(workload: Workload, opts: Options, iterations: usize) -> anyhow::Result<()> {
```
djc, line 319: I think it might make sense to store `iterations` inside `Options`?

src/lib.rs
```diff
@@ -328,23 +328,25 @@ pub fn generate_flamegraph_for_workload(workload: Workload, opts: Options) -> an
...
-            opts.frequency,
-            opts.custom_cmd,
-            opts.verbose,
-            opts.ignore_status,
-        )
-    };
-
+    for _i in 0..iterations {
+        if let Workload::ReadPerf(_perf_file) = &workload {
+            // pass
```
djc, line 333: Did you test this with a `Workload::ReadPerf` scenario? I guess maybe this `if` needs to be outside this loop, so that we can still pass `perf_file` to `output()` (as `perf_output`)?

djc (commented):
Looks like CI isn't passing yet, so that needs some work.

dvdplm:
@djc I am interested in this feature as well, but maybe @bartlomieju got busy in the meantime. Would it be ok if I finished up the PR and got it over the hump?

djc:
@dvdplm sounds good to me. Make sure to retain @bartlomieju as a (co-)author of the commits, and probably good to just squash all commits into a single one.

bartlomieju:
Thanks for picking this up, I completely forgot about this PR. I'm gonna close it for now.

PR #276, opened 2023-08-05
Add support for profiling unit-test style benchmarks

d-e-s-o:
This change extends the program with support for profiling unit test
benchmarks, aka libtest benchmarks. Support is accomplished in a vein
similar to that of unit tests: we introduce the `--unit-bench` argument
that behaves very much like the existing `--unit-test` does, but invokes
benchmarks instead of tests.
We cannot piggy-back on the existing `--bench` argument for the same
reason (presumably) that `--unit-test` is separate from `--test`: `--bench`
and `--test` require a target to be provided, but `--unit-test` and
`--unit-bench` can work without an argument or can be a pattern of tests
to match against, but not a target.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -38,6 +38,12 @@ struct Opt {
     #[clap(long, group = "exec-args")]
     unit_test: Option<Option<String>>,
 
+    /// Crate target to unit benchmark, <bench> may be omitted if crate only has one target
+    /// (currently profiles the test harness and all tests in the binary; test selection
+    /// can be passed as trailing arguments after `--` as separator)
+    #[clap(long, group = "exec-args")]
+    unit_bench: Option<Option<String>>,
+
     /// Benchmark to run
     #[clap(long, group = "exec-args")]
     bench: Option<String>,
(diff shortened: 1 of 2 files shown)


src/bin/cargo-flamegraph.rs
```diff
@@ -183,6 +198,13 @@ fn workload(opt: &Opt, artifacts: &[Artifact]) -> anyhow::Result<Vec<String>> {
             unit_test: Some(Some(t)),
             ..
         } => (&["lib", "bin"], t),
+        Opt {
+            unit_bench: Some(Some(t)),
+            ..
+        } => {
+            trailing_arguments.push("--bench".to_string());
```
djc, line 205: What is this trying to achieve? It seems like a weird setup to append to a copy of `trailing_arguments`.

djc (commented):
Are unit-test benchmarks a documented concept in Cargo?

d-e-s-o:
> Are unit-test benchmarks a documented concept in Cargo?

[Yes](https://doc.rust-lang.org/1.4.0/book/benchmark-tests.html).

src/bin/cargo-flamegraph.rs
d-e-s-o: One runs benchmarks with the `--bench` switch provided to the unit test binary. See `<binary> --help`. We were already doing a copy a few lines down, so this doesn't really change much.

djc (approved):
Alright, thanks for the explanations!

PR #290, opened 2023-09-22
make --unit-test and --unit-bench target library by default

### LINKED ISSUE #245: --unit-test doesnt seem to be able to prioritise lib over bin
Using --unit-test when my crate builds both a single bin and a single lib fails.

$ cargo flamegraph --unit-test -- my_test 
Error: several possible targets found: [BinaryTarget { package: "odonata", target: "odonata", kind: ["lib"] }, BinaryTarget { package: "odonata", target: "odonata", kind: ["bin"] }], please pass an explicit target.

But --unit-test and --bin are incompatible options (and really I want to specify lib somehow)

--verbose doesnt show any additional info

Removing "bin" from the below fixes this - but probably isnt the right fix.

> 
> } else if let Some(unit_test) = opt.unit_test {
>         let target = find_unique_target(
>             &["bin", "lib"],
>             opt.package.as_deref(),
>             opt.manifest_path.as_deref(),
>             unit_test.as_deref(),
>         )?;
>

carloskiki:
This is an attempt to resolve #245 as it makes it possible to target the library or binary of a crate with identical library and binary names.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -36,14 +36,16 @@ struct Opt {
     #[clap(long, group = "exec-args")]
     test: Option<String>,
 
-    /// Crate target to unit test, <unit-test> may be omitted if crate only has one target
-    /// (currently profiles the test harness and all tests in the binary; test selection
+    /// Crate target to unit test, <unit-test> may be omitted if the crate's library is targeted. To
+    /// target a binary, it name must be specifed.
+    /// (currently profiles the test harness and all tests in the library/binary; test selection
     /// can be passed as trailing arguments after `--` as separator)
     #[clap(long, group = "exec-args")]
     unit_test: Option<Option<String>>,
 
-    /// Crate target to unit benchmark, <bench> may be omitted if crate only has one target
-    /// (currently profiles the test harness and all tests in the binary; test selection
+    /// Crate target to unit benchmark, <bench> may be omitted if the crate's library is targeted. To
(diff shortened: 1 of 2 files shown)


src/bin/cargo-flamegraph.rs
```diff
@@ -36,14 +36,16 @@ struct Opt {
     #[clap(long, group = "exec-args")]
     test: Option<String>,
 
-    /// Crate target to unit test, <unit-test> may be omitted if crate only has one target
-    /// (currently profiles the test harness and all tests in the binary; test selection
+    /// Crate target to unit test, <unit-test> may be omitted if the crate's library is targeted. To
```
djc, line 39: Can we preserve the property that we test whatever the target is if there is a single one? Seems silly to require that we specify the bin target if there is a single target.

Carlosted: I know _this is silly_. The only way to keep the old behaviour (_I think_) requires major refactoring regarding the `find_single_target` function. If there are any future refactoring plans, I am up to the challenge.

Otherwise this is the best way to hot fix it.

Also to add to the explanation of #245, this is quite a major bug/missing feature, as it is really frequent to have a binary in `main.rs` that relies on the code of its library in `lib.rs`. There is currently no way of distinguishing between the binary and the library in this type of situation.

djc: I don't want this hotfix. Why is the required refactoring so "major"?

Issue #47, opened 2019-10-31
problem with envionment variables 


zhangli-pear:
My program work with envionment variables, but it's seems like it cann't find them when I run flamegraph.

so how can i pass all the envionment variables to my program?

agodnic:
I'm having the same issue.

jonas-schievink:
This happens with `--root` only, since `sudo` doesn't pass through environment variables by default. I'm not sure if doing that by default is a good idea though (it's not the default for a reason).

zhangli-pear:
> This happens with `--root` only, since `sudo` doesn't pass through environment variables by default. I'm not sure if doing that by default is a good idea though (it's not the default for a reason).

`root` is required to run the command.
Is that possible to add an option to allow passing envionment variables explicitly.

alextes:
Are maintainers open to a PR?

Do people prefer:
* an extra flag to pass?
* always passing with `--root`?

SichangHe:
The question I have is: how to pass through environment variables at all?
I am using `env-logger` and cannot get my logs without hardcoding a logging level.

SichangHe:
> The question I have is: how to pass through environment variables at all? I am using `env-logger` and cannot get my logs without hardcoding a logging level.

Okay. Running with `sudo --preserve-env=…` without `--root` works.

PR #295, opened 2023-11-24
Optional sudo flags as --root argument

Korrigan:
Thank you for this useful tool! :)

This PR allows to configure the sudo behavior when `--root` is used so one can provide additional command-line flags which will be passed to "sudo".
This is especially useful when preserving environment is needed: `cargo flamegraph --root='--preserve-env' --bin my-bin` (so this provides a solution for #47 ).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,7 +148,7 @@ mod arch {
 
     pub(crate) fn initial_command(
         workload: Workload,
-        sudo: bool,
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

larseggert:
Any chance for a release that would ship this PR?

djc:
Published 0.6.5.

PR #296, opened 2023-12-29
Update inferno dependency to 0.11.19

SergejIsbrecht:
Update to latest inferno version to include 0.11.12 in order to increase flamegraph.svg rendering performance

https://github.com/jonhoo/inferno/pull/262


djc:
Since this is a semver-compatible version bump, there really is no need to change `Cargo.toml`. Also, using `cargo install` to install the `flamegraph`/`cargo-flamegraph` binaries off of crates.io should already be using latest inferno 0.11.x without this change, so I don't think this a very useful change.

SergejIsbrecht:
@djc , I have to cross compile for aarch64, therefore I can not use `cargo install`

PR #298, opened 2024-01-06
Run perf script as root to resolve kallsyms

### LINKED ISSUE #297: Running with `--root` cannot read kallsyms
Currently:

* https://github.com/flamegraph-rs/flamegraph/blob/v0.6.4/src/lib.rs#L102-L104

> We executed `perf record` with sudo, and will be executing `perf script` *without* sudo.

Running as a regular user means that one cannot read `/proc/kallsyms` to resolve kernel symbols, resulting in this:

image

When running `perf script`:

```
rs-1brc  5713  7909.014673:    1003009 cpu-clock:pppH:
        ffffffe0d6f5a7e0 [unknown] ([kernel.kallsyms])
        ffffffe0d6f5b290 [unknown] ([kernel.kallsyms])
        ffffffe0d6611558 [unknown] ([kernel.kallsyms])
                    6f4c core::intrinsics::copy_nonoverlapping+0x4a4 (inlined)
```

When running `sudo perf script -f`:

```
rs-1brc  5713  7909.014673:    1003009 cpu-clock:pppH:
        ffffffe0d6f5a7e0 el0_da+0x28 ([kernel.kallsyms])
        ffffffe0d6f5b290 el0t_64_sync_handler+0xc0 ([kernel.kallsyms])
        ffffffe0d6611558 el0t_64_sync+0x198 ([kernel.kallsyms])
                    6f4c core::intrinsics::copy_nonoverlapping+0x4a4 (inlined)
```

The latter is definitely preferred from a users point of view.

bobrik:
Closes #297.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,24 +41,14 @@ mod arch {
 
     pub(crate) fn initial_command(
         workload: Workload,
-        sudo: Option<Option<String>>,
+        sudo: Option<Option<&str>>,
         freq: Option<u32>,
         custom_cmd: Option<String>,
         verbose: bool,
         ignore_status: bool,
     ) -> Option<String> {
         let perf = env::var("PERF").unwrap_or_else(|_| "perf".to_string());
-
-        let mut command = if let Some(sudo) = sudo {
-            let mut c = Command::new("sudo");
-            if let Some(sudo_args) = sudo {
-                c.arg(sudo_args);
-            }
-            c.arg(perf);
-            c
-        } else {
-            Command::new(perf)
-        };
+        let mut command = sudo_command(&perf, sudo);
 
         let args = custom_cmd.unwrap_or(format!(
             "record -F {} --call-graph dwarf,16384 -g",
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -41,24 +41,15 @@ mod arch {
 
     pub(crate) fn initial_command(
         workload: Workload,
-        sudo: Option<Option<String>>,
+        sudo: Option<&Option<String>>,
```
djc, line 44: I'd prefer to change the type of this argument to `Option<Option<&str>>`.

src/lib.rs
```diff
@@ -99,26 +90,19 @@ mod arch {
     pub fn output(
         perf_output: Option<String>,
         script_no_inline: bool,
-        sudo: bool,
+        sudo: Option<&Option<String>>,
```
djc, line 93: Same here.

src/lib.rs
```diff
@@ -99,26 +90,19 @@ mod arch {
...
-                    .wait()
-                    .expect(arch::WAIT_ERROR);
-            }
-        }
-
+        // We executed `perf record` with sudo, and will be executing `perf script` with sudo,
+        // so that we can resolve privileged kernel symbols from /proc/kallsyms.
         let perf = env::var("PERF").unwrap_or_else(|_| "perf".to_string());
-        let mut command = Command::new(perf);
+
```
djc, line 98: Nit: please remove the extra empty line.

src/lib.rs
```diff
@@ -151,24 +135,15 @@ mod arch {
 
     pub(crate) fn initial_command(
         workload: Workload,
-        sudo: Option<Option<String>>,
+        sudo: Option<&Option<String>>,
```
djc, line 138: And here.

src/lib.rs
```diff
@@ -292,6 +267,19 @@ mod arch {
     }
 }
 
+fn sudo_command(sudo: Option<&Option<String>>, command: &str) -> Command {
```
djc, line 270: I'd like to see the following changes:

* Invert the order of the arguments
* Change the logic inside to do an early return for the simple case, reducing rightward drift:
  ```rust
  let sudo = match sudo {
    Some(sudo) => sudo,
    None => return Command::new(command),
  };
  
  ..
  ```

src/lib.rs
```diff
@@ -338,13 +326,12 @@ pub fn generate_flamegraph_for_workload(workload: Workload, opts: Options) -> an
...
     };
 
-    let use_sudo = opts.root.is_some();
     let perf_output = if let Workload::ReadPerf(perf_file) = workload {
         Some(perf_file)
     } else {
         arch::initial_command(
             workload,
-            opts.root,
+            opts.root.as_ref(),
```
djc, line 334: Because we'll need a more complicated transformation here, let's extract it at the point where `use_sudo` was previously defined so we can pass the output to both `initial_command()` and `output()`?

djc (commented):
This looks pretty good, just have some stylistic nits to address.

bobrik:
Updated to address the comments.

src/lib.rs
```diff
@@ -338,13 +324,17 @@ pub fn generate_flamegraph_for_workload(workload: Workload, opts: Options) -> an
         signal_hook::low_level::register(SIGINT, || {}).expect("cannot register signal handler")
     };
 
-    let use_sudo = opts.root.is_some();
+    let sudo = opts
+        .root
+        .as_ref()
+        .map(|inner| inner.as_ref().map(String::as_str));
```
djc, line 330: Can we use `as_deref()` here instead of the `as_ref().map(..)` construction?

djc (approved):
One more question, happy to merge this soon!

src/lib.rs
bobrik: Sure thing, updated.

djc:
Thanks!

Issue #229, opened 2022-09-29
Can't build after running flamegraph (MacOS)


roxgib:
After running `sudo cargo flamegraph` (which runs fine) I get one of the following two errors from `cargo build --release`:
```
error: linking with `cc` failed: exit status: 1
[_massive list of files omitted_]
 = note: ld: can't write output file: ~/project/target/release/deps/project-47ba9299aacab074 for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: could not compile `project` due to previous error; 
```
```
error: failed to open: ~project/target/release/.cargo-lock
Caused by:
  Permission denied (os error 13)
```
But running `sudo cargo flamegraph` again works fine. I assume it's leaving a lock on some files which prevents access. `cargo clean` hits the same problem, so I have to manually delete the target directory and rebuild. Might be related to using sudo, but unfortunately this seems to be required on MacOS.

djc:
Yeah, I'm afraid you just have to `sudo cargo clean` before you rebuild without `sudo`, and I don't think there's an easy way around that. PRs welcome for improving the documentation around this, though!

popematt:
Another workaround is to run `cargo build` with the same build options that you are going to use for `cargo flamegraph` immediately before running `sudo cargo flamegraph`. That way nothing needs to be rebuilt while running `sudo cargo flamegraph`.

@djc Just so that I'm understanding it correctly, the _only_ thing that needs to run as root is running dtrace. Building as root and writing the final flamegraph as root are unintended side-effects, right?

djc:
> @djc Just so that I'm understanding it correctly, the _only_ thing that needs to run as root is running dtrace. Building as root and writing the final flamegraph as root are unintended side-effects, right?

I think so!

zbentley:
The `--root` argument to `cargo flamegraph` avoids this issue. If `cargo` itself is run with `sudo`, all of its bookkeeping side effects (like the lockfile) before running `flamegraph` are also run as root, which breaks things.

Instead, the README-recommended invocation of e.g. `cargo flamegraph --root --bin somebin` will cause `flamegraph` to internally use `sudo` only where it is required, after `cargo` has finished its build.

PR #303, opened 2024-02-03
Address #229: Improve documentation and error output on MacOS when not running as root

zbentley:
Addresses/mitigates #229 

Two changes in this PR: 
- Force an error when running via `cargo flamegraph` on MacOS without the `--root` flag. 
- Update documentation regarding the reasons for preferring the `--root` flag over `sudo cargo`.

Additional things that I could do if needed:
- Add equivalent runtime errors to the non-`cargo` standalone `flamegraph` binary. I didn't do that because dtrace will indicate an error anyway without permissions, and `sudo flamegraph` isn't as much of a problem as `sudo cargo`. 
- Force an error if euid==0 in the `cargo flamegraph` binary, telling people not to run cargo as root. That seems overly perscriptive as there are probably some folks who intentionally do this.
- Actually dry-run dtrace fully against a dummy program, intercept error output, and provide suggestions based on the presence of known errors. This would potentially be useful (and not just on MacOS) but would be a lot more work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ clap = { version = "4.0.11", features = ["derive"] }
 clap_complete = "4.0.2"
 inferno = { version = "0.11.0", default_features = false, features = ["multithreaded", "nameattr"] }
 opener = "0.6.0"
-shlex = "1.3.0"
+shlex = "1.1.0"
 
 [target.'cfg(unix)'.dependencies]
 signal-hook = "0.3.10"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ clap = { version = "4.0.11", features = ["derive"] }
 clap_complete = "4.0.2"
 inferno = { version = "0.11.0", default_features = false, features = ["multithreaded", "nameattr"] }
 opener = "0.6.0"
-shlex = "1.1.0"
+shlex = "1.3.0"
 
 [target.'cfg(unix)'.dependencies]
 signal-hook = "0.3.10"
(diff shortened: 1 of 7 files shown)


djc (approved):
Thanks!

zbentley:
@djc you want me to bump shlex to resolve that audit advisory build failure?

README.md
```diff
@@ -182,8 +182,10 @@ echo -1 | sudo tee /proc/sys/kernel/perf_event_paranoid
 ### DTrace on macOS
 
 On macOS, there is no alternative to running as superuser in order to
-enable dtrace. The simplest way is to use `--root`, this way `rustc`
-will be run normally but the binary will get run as superuser.
+enable DTrace. This should be done by invoking `cargo flamegraph --root ...`.
```
djc, line 185: I think this is overindexing on using `cargo flamegraph`, when just `flamegraph` is also perfectly viable.

Cargo.toml
```diff
@@ -25,7 +25,7 @@ clap = { version = "4.0.11", features = ["derive"] }
 clap_complete = "4.0.2"
 inferno = { version = "0.11.0", default_features = false, features = ["multithreaded", "nameattr"] }
 opener = "0.6.0"
-shlex = "1.1.0"
+shlex = "1.3.0"
```
djc, line 28: Please remove this -- the requirements have not changed and I've updated the lockfile to use 1.3.0 already.

README.md
zbentley: Updated!

djc (approved):

PR #309, opened 2024-03-06
Improve documentation for frequency arg

### LINKED ISSUE #307: Frequency option is hardly documented
From the `--help` output, it's impossible to tell

- What kind of input format is accepted (apparently the input is a number in Hz)
- What the default value is (AFAIU this is determined by perf's default, which _appears to be_ 1000 Hz, on my system)

jplatte:
Hope you don't mind the drive-by code improvements 😉

Resolves #307.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -478,7 +478,7 @@ pub struct Options {
     #[clap(long, value_name = "SUDO FLAGS")]
     pub root: Option<Option<String>>,
 
-    /// Sampling frequency in Hz
+    /// Sampling frequency in Hz [default: 997]
     #[clap(short = 'F', long = "freq")]
     frequency: Option<u32>,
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -478,7 +478,7 @@ pub struct Options {
     #[clap(long, value_name = "SUDO FLAGS")]
     pub root: Option<Option<String>>,
 
-    /// Sampling frequency in Hz
+    /// Sampling frequency in Hz [default: 997]
```
djc, line 477: I think we could also set `default = "997"` in the `clap` attribute, which I think would allow changing the type here to `u32` and would probably take care of the docstring for us?

djc (approved):
I love me some drive-by improvements!

src/lib.rs
jplatte: I wanted to do that but `fn check` below checks whether the argument is specified explicitly which that would break.

djc:
Thanks!

PR #322, opened 2024-05-15
Add unit-test-kind option to allow choosing bin or lib

felipou:
When running with the unit-test option, there was no way to choose between a bin or lib target, so the result was always an error if there were both types for the same target name. To allow running unit tests in that case, this new option unit-test-kind may be used to specify the type of target that we want.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -6,6 +6,13 @@ use clap::{Args, Parser};
 
 use flamegraph::Workload;
 
+#[derive(Debug, Clone, Copy, clap::ValueEnum)]
+#[clap(rename_all = "snake_case")]
+enum UnitTestTargetKind {
+    Bin,
+    Lib,
+}
+
 #[derive(Args, Debug)]
 struct Opt {
     /// Build with the dev profile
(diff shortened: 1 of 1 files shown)


src/bin/cargo-flamegraph.rs
```diff
@@ -436,8 +448,14 @@ fn main() -> anyhow::Result<()> {
         opt.package = Some(target.package);
         target.kind
     } else if let Some(unit_test) = opt.unit_test {
+        let kinds: &[&str] = match opt.unit_test_kind {
```
djc, line 451: Nit: I don't think we need the type annotation here? Would prefer to do without it.

djc (approved):
This seems reasonable, thanks!

src/bin/cargo-flamegraph.rs
felipou: I always prefer to avoid those too, but when I first tried without it, I got the following error:
```
error[E0308]: `match` arms have incompatible types
   --> src/bin/cargo-flamegraph.rs:454:21
    |
451 |           let kinds = match opt.unit_test_kind {
    |  _____________________-
452 | |             Some(UnitTestTargetKind::Bin) => &["bin"],
    | |                                              -------- this is found to be of type `&[&str; 1]`
453 | |             Some(UnitTestTargetKind::Lib) => &["lib"],
    | |                                              -------- this is found to be of type `&[&str; 1]`
454 | |             None => &["bin", "lib"],
    | |                     ^^^^^^^^^^^^^^^ expected an array with a fixed size of 1 element, found one with 2 elements
455 | |         };
    | |_________- `match` arms have incompatible types
```

felipou: Yes! Should I push that change? Also, should I add a comment about this on that line? Seems a little "cryptic" to me, I probably wouldn't understand why the `[..]` is there just by looking at the code :sweat_smile:

djc: Comment is fine if you think it helps. Please squash this change into the existing commit.

felipou: Done!

Issue #223, opened 2022-09-08
Allow setting title


casey:
It would be nice to be able to change the titles of generated flamegraphs to something other than Flame Graph. Maybe with a `--title STRING` option.

djc:
This seems sort of the same thing as #222 -- do you really need both? In order to implement these in flamegraph we'd need API exposed by the inferno crate which renders the SVG for us. flamegraph is really just the glue between running perf (or other similar tools on other platforms) and pushing that into inferno.

casey:
Ideally I'd like to be able to do both. I want to set a short, concise title that replaces the current "Flame Graph" title, and also include some notes. The title would be something like "Foo Bar Benchmark 500000" and the notes would be longer. "Built from commit xyz, using a b and c parameters.

djc:
Okay, just to set expectations, I'm unlikely to write code for this myself, but happy to provide you with guidance on how to glue CLI to inferno -- probably need to start figuring out if inferno has support for this stuff.

casey:
No worries! I'm pretty backed up, so I'm unlikely to get to this for a while.

casey:
No worries! I'm pretty backed up, so I'm unlikely to get to this for a while.

PR #326, opened 2024-06-17
Add ability to customize title and subtitle of generated graph

jamel:
As suggested in #223 it would be nice to support `--title` command line argument to set graph title in produced SVG image. To make flamegraph even more compatible with original flamegraph.pl here `--subtitle` command line argument is added as well.


djc (approved):

djc:
Thanks!

PR #327, opened 2024-06-25
Update README.md to mention 'mold'

ChrisJefferson:
Same issue with 'lld' also occurs with 'mold'.

I realise how you use mold at all looks a bit weird, it comes from https://stackoverflow.com/questions/67511990/how-to-use-the-mold-linker-with-cargo (just for information)


djc (approved):
Thanks!

PR #334, opened 2024-08-12
Add kernel, global, and timeout flags

mkeeter:
We've been using `flamegraph` for full-system benchmarking at [Oxide](https://oxide.computer), relying heavily on custom sampling commands (since it only samples one process by default).  This PR would let us stop using custom commands by adding their functionality into the main `flamegraph` executable:

- `--kernel` changes the flamegraph probe to also capture kernel stack frames
- `--global` captures a flamegraph of everything in the system, instead of a single process
- `--timeout` tells `flamegraph` to exit after a certain amount of time, e.g. to capture 1 second of full-system traces

The results are pretty neat:

fg4

I have only implemented these flags for the DTrace backend, but `perf` will print helpful error messages if you try to use them.  If you want to maintain strict CLI parity, let me know and I can dig into what `perf` changes would be required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -491,6 +491,7 @@ dependencies = [
  "cargo_metadata",
  "clap",
  "clap_complete",
+ "humantime",
  "inferno",
  "opener",
  "shlex",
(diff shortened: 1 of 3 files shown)


djc:
> If you want to maintain strict CLI parity, let me know and I can dig into what perf changes would be required.

Yeah, would be nice if you could spend at least a little time on this.

Issue #270, opened 2023-07-06
[Feature Requst] Show Progress


lixin-wei:
In my case, `perf.data` is huge(10GB+) and usually takes 30+ min to generate the flame graph.

It will be much more friendly if there is a progress bar.

djc:
Unfortunately I don't have much time for maintaining flamegraph. Maybe consider submitting a PR?

djc:
No, but it's pretty simple. We get the inferno crate to collapse the data at https://github.com/flamegraph-rs/flamegraph/blob/main/src/lib.rs#L367 and generate the flamegraph at https://github.com/flamegraph-rs/flamegraph/blob/main/src/lib.rs#L427. You should figure which of those is slow. Then, I think you might want to wrap an indicatif thing around the input to either of those things.

smoelius:
@djc Do these look like the right links? I'm speculating based on the date of your comment:
- https://github.com/flamegraph-rs/flamegraph/blob/1aeb2a82cbe228d715f38751fd7e1a5c3586dfd4/src/lib.rs#L367
- https://github.com/flamegraph-rs/flamegraph/blob/1aeb2a82cbe228d715f38751fd7e1a5c3586dfd4/src/lib.rs#L427

smoelius:
Also, for me, the bottleneck appears to be `perf script`: https://github.com/flamegraph-rs/flamegraph/blob/1aeb2a82cbe228d715f38751fd7e1a5c3586dfd4/src/lib.rs#L128

Is this surprising to you? (I don't know what this command does.)

gth828r:
This is a problem which I had, so I can probably take a stab at adding this feature. Implementing a true progress bar with perf script is not obvious to me, but I think a spinner with a message about what is running and an elapsed time is a good start, and that is fairly simple to add.

gth828r:
Ok, after a few different iterations, I found something that I like and it is available as a PR at https://github.com/flamegraph-rs/flamegraph/pull/343

PR #343, opened 2024-11-07
Show indicators that perf script is running

gth828r:
Perf script can take a long time to run. This change adds an indicator to show that it is running so that the user knows that their application is not causing the terminal to hang.

Addresses #270

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,6 +34,9 @@ pub enum Workload {
 
 #[cfg(target_os = "linux")]
 mod arch {
+    use indicatif::{ProgressBar, ProgressStyle};
+    use std::time::Duration;
+
     use super::*;
 
     pub const SPAWN_ERROR: &str = "could not spawn perf";
@@ -116,7 +119,18 @@ mod arch {
             command.arg(perf_output);
         }
 
-        let output = command.output().context("unable to call perf script")?;
+        // perf script can take a long time to run. Notify the user that it is running
+        // by using a spinner. Note that if this function exits before calling
+        // spinner.finish(), then the spinner will be completely removed from the terminal.
+        let spinner = ProgressBar::new_spinner().with_prefix("Running perf script");
+        spinner.set_style(
+            ProgressStyle::with_template("{prefix} [{elapsed}]: {spinner:.green}").unwrap(),
+        );
+        spinner.enable_steady_tick(Duration::from_millis(500));
+
+        let result = command.output().context("unable to call perf script");
+        spinner.finish();
+        let output = result?;
         if !output.status.success() {
             anyhow::bail!(format!(
                 "unable to run 'perf script': ({}) {}",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -350,6 +350,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b63caa9aa9397e2d9480a9b13673856c78d8ac123288526c37d7839f2a86990"
 
 [[package]]
+name = "console"
+version = "0.15.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0e1f83fc076bd6dd27517eacdf25fef6c4dfe5f1d7448bafaaf3a26f13b5e4eb"
+dependencies = [
+ "encode_unicode",
+ "lazy_static",
+ "libc",
+ "unicode-width",
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "core-foundation"
 version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -464,6 +477,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "encode_unicode"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a357d28ed41a50f9c765dbfe56cbc04a64e53e5fc58ba79fbc34c10ef3df831f"
+
+[[package]]
 name = "encoding_rs"
 version = "0.8.35"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -493,6 +512,7 @@ dependencies = [
  "cargo_metadata",
  "clap",
  "clap_complete",
+ "indicatif",
  "inferno",
  "opener",
  "shlex",
@@ -851,6 +871,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "indicatif"
+version = "0.17.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "763a5a8f45087d6bcea4222e7b72c291a054edf80e4ef6efd2a4979878c7bea3"
+dependencies = [
+ "console",
+ "instant",
+ "number_prefix",
+ "portable-atomic",
+ "unicode-width",
+]
+
+[[package]]
 name = "inferno"
 version = "0.11.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


gth828r:
I want to make a quick tweak and see if I can improve the case where the user cancels the run. Right now, it always appears that a run is successful from the output of the spinner. That is OK, but I think we can do better. The tradeoff is that the code will be a little bit more complex.

gth828r:
> I want to make a quick tweak and see if I can improve the case where the user cancels the run. Right now, it always appears that a run is successful from the output of the spinner. That is OK, but I think we can do better. The tradeoff is that the code will be a little bit more complex.

This attempted change did not work. perf script will return a successful status code even if it was forcefully terminated by a signal. I just reverted this change, because the original code was much simpler and cleaner. Turning off draft status.

gth828r:
A quick note on the approach:

The main goal of this change is that under normal operations, users should see that flamegraph-rs is doing work when it calls `perf script`, which can take a long time. I do this with a spinner, because I am not sure how detect progress over time.

As a secondary goal, I had hoped that the final output of the spinner could reflect success or failure, particularly if the command was terminated early with a signal. Unfortunately, when testing I found that if a signal is used to stop `perf script` early, the process output status code is still success, so I am not sure we have a way to tell that the process was terminated early. Users just need to be aware of whether or not they terminated things when they were still running -- not much can be done about it unless the behavior of `perf script` changes.

With this constraint in mind, I am using a scopeguard block to always mark the spinner as finished, regardless of how the function exits. This helps keep the spinner code isolated and simple while also ensuring that the spinner resources are always cleaned up, even if the function exits early with an error.

src/lib.rs
```diff
@@ -34,6 +34,10 @@ pub enum Workload {
 
 #[cfg(target_os = "linux")]
 mod arch {
+    use indicatif::{ProgressBar, ProgressStyle};
+    use scopeguard;
```
smoelius, line 38:
```suggestion
```
I think this line may be unnecessary?

djc:
(If you rebase on current main, the CI audit failure should go away).

src/lib.rs
```diff
@@ -116,6 +120,20 @@ mod arch {
...
+        // by using a spinner. Ensure that the spinner gets cleaned up regardless of how
+        // this function exits by using a scopeguard.
+        let spinner = ProgressBar::new_spinner().with_prefix("Running perf script");
+        spinner.set_style(
+            ProgressStyle::with_template("{prefix} [{elapsed}]: {spinner:.green}")
+                .unwrap()
+                .tick_strings(&[".", "..", "...", " ..", "  .", "", "✓"]),
+        );
+        spinner.enable_steady_tick(Duration::from_millis(500));
+        let _guard = scopeguard::guard(spinner, |spinner| {
```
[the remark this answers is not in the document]
gth828r: I had been concerned that if the following exited the function with an error without calling "finish" on the spinner, that the spinner might not be properly cleaned up and could clutter the terminal.
```rust
let output = command.output().context("unable to call perf script")?;
```
In reality, I just tried it and if you don't call "finish", it just completely removes the spinner from the terminal output altogether, which is reasonable. I will go ahead and do as you suggest. Given what I know now, I agree that scopeguard is not necessary.

src/lib.rs
gth828r: Thanks! This is OBE since we removed scopeguard altogether, but you were right.

gth828r:
Rebased on main and removed the scopeguard dependency.

src/lib.rs
```diff
@@ -116,7 +119,19 @@ mod arch {
...
+        let spinner = ProgressBar::new_spinner().with_prefix("Running perf script");
+        spinner.set_style(
+            ProgressStyle::with_template("{prefix} [{elapsed}]: {spinner:.green}")
+                .unwrap()
+                .tick_strings(&[".", "..", "...", " ..", "  .", "", "✓"]),
+        );
+        spinner.enable_steady_tick(Duration::from_millis(500));
+
         let output = command.output().context("unable to call perf script")?;
+        spinner.finish();
```
djc, line 134: So we can more cleanly always call `spinner.finish()` after `command.output()` (before yielding the error using the `?` operator). I think I would prefer that.

src/lib.rs
```diff
@@ -116,7 +119,19 @@ mod arch {
...
         }
 
+        // perf script can take a long time to run. Notify the user that it is running
+        // by using a spinner. Note that if this function exits before calling
+        // spinner.finish(), then the spinner will be completely removed from the terminal.
+        let spinner = ProgressBar::new_spinner().with_prefix("Running perf script");
+        spinner.set_style(
+            ProgressStyle::with_template("{prefix} [{elapsed}]: {spinner:.green}")
+                .unwrap()
+                .tick_strings(&[".", "..", "...", " ..", "  .", "", "✓"]),
```
djc, line 129: Why deviate from the default style?

gth828r: Let's split this into two parts: the template and the spinner.

In terms of changing the template, I think there are some important deviations from the default:
 1. I don't believe there is a way to show the elapsed time in the default style template. I tried calling `with_elapsed(Duration::from_millis(0))` on the progress bar instantiation using the default style, and I don't see anything. I'm not a power user of indicatif, so maybe I am doing it wrong, but setting the style with the template seemed like the best way to include the elapsed time. **I feel strongly** that including the elapsed time is important since I cannot find a way to implement a true progress bar. If there are better ways to include the elapsed time, we can use those.
 2. The default style does not include a prefix that comes before the spinner. You can include a message which comes after the spinner, but I personally think that looks strange. I am used to seeing something of the form `<what is happening>: <status>`. I **don't** feel as strongly about this one, so if you prefer the message after the spinner that is fine.

Given that, please let me know how you want to proceed with changing the template.


For the spinner itself, we can switch back to the default tick strings. I personally didn't like it in part because the bottom of the spinner doesn't line up with the bottom of the prompt, but any style or color works. Shall I go ahead and switch back to the default spinner?

src/lib.rs
gth828r: The behavior here will depend on what happens with our discussion down at https://github.com/flamegraph-rs/flamegraph/pull/343#discussion_r1835270990, so let me just lay out some details. I don't really feel strongly about any of this, but let's resolve that discussion first before we try to resolve this one. Here are some details about why I want to do these in that particular order:

I've been playing around with different scenarios to see what happens in different situations with the spinner. What I have found is that the resources are always cleaned up when the spinner goes out of scope. The only thing that calling `spinner.finish()` seems to do is leave a copy of the prefix/message/final tick in the terminal history. Note that the final tick string is configurable, and it is only called once finish() gets set (not part of the spinner animation), so using a special final tick symbol can act as an indicator of success.

If you don't call `spinner.finish()` before the spinner goes out of scope, then the line where the spinner was is effectively removed from the terminal history, as if the spinner wasn't ever there. If you call finish whether the command succeeds or fails, then we will see the terminal history including the final tick. The one exception is that if you don't use a message/prefix and the default tick is blank, then nothing will be included in the terminal history.

With all of this in mind, if you use the default spinner, the final tick of the spinner is simply blank, so it doesn't really indicate anything. My custom ticks use a checkmark to indicate successful completion. The effect of doing what you propose with the default spinner would be to always have the spinner line left in the terminal history regardless of whether the command succeeded or failed. I think that is OK. If we use a custom final tick, then I think this change would be confusing, because the user might think the command succeeded. If we actually hit the error case here, then in reality it means that the process was never started up in the first place, so I felt that just removing the spinner message from the terminal history made sense.

src/lib.rs
djc: I think I'm fine with the template, was mostly responding to the custom spinner. Let's remove that for now? If you think the default spinner is bad, consider advocating a change upstream (where I'm also a maintainer).

src/lib.rs
djc: Right, I still think it makes sense to unconditionally call `.finish()` so we leave the elapsed time in terminal history, which I think will be useful in both the success and error cases.

djc:
Thanks, this looks good to me. The audit failure in CI is unrelated.

src/lib.rs
gth828r: > If you think the default spinner is bad, consider advocating a change upstream (where I'm also a maintainer).

I wouldn't say "bad", just the bottom of the spinner didn't line up with the bottom of my prefix text. No big deal, and possibly even specific to how my system is rendering it. Like I said before, I'm not a good person to consult on style decisions anyway!

PR #350, opened 2024-12-03
docs: add slow `addr2line` solution section

alan910127:
## Changes

This PR adds a new section to the README.md titled "Use custom `addr2line` binary for perf," which provides a simple solution to address the issue of slow `perf` command execution. The solution involves using [gimli-rs/addr2line](https://github.com/gimli-rs/addr2line) as an alternative to the default `addr2line` binary.


djc (approved):
Thanks, looks great!

Issue #248, opened 2023-01-16
PERFILE binary contents spewed on stdout on Github Actions runners?


brainstorm:
I'm running `cargo flamegraph --deterministic --flamechart -o flamechart.svg --bench search-benchmarks -- --bench` on Github Actions and [I'm seeing the binary contents of perf files being dumped on stdout (perhaps stderr?)](https://github.com/umccr/htsget-rs/actions/runs/3934614655/jobs/6729521082):

![Screenshot 2023-01-17 at 10 44 51 am](https://user-images.githubusercontent.com/175587/212781107-48dd561a-06a4-44d7-9fb1-36d26d988f3f.png)

The same command on OSX (DTrace) does not seem to have that issue.

djc:
Seems surprising -- I don't have an idea off the top of my head why it might be.

brainstorm:
I just tested it under WSL2 instead of GHA and the output both surprised me and gave some clues... 3.6GB of `perf.data`!!:

```
brainstorm@braingate:~/htsget-rs$ cargo flamegraph --deterministic --flamechart -o flamechart.svg --bench search-benchmarks -- --bench
    Finished bench [optimized + debuginfo] target(s) in 0.50s
Queries/[LIGHT] Bam query all
                        time:   [601.06 ns 607.58 ns 615.31 ns]
Found 5 outliers among 50 measurements (10.00%)
  1 (2.00%) low mild
  2 (4.00%) high mild
  2 (4.00%) high severe
Queries/[LIGHT] Bam query specific
                        time:   [606.72 ns 620.28 ns 636.32 ns]
Found 7 outliers among 50 measurements (14.00%)
  1 (2.00%) high mild
  6 (12.00%) high severe
Queries/[LIGHT] Bam query header
                        time:   [597.83 ns 600.43 ns 603.43 ns]
Found 7 outliers among 50 measurements (14.00%)
  2 (4.00%) low mild
  3 (6.00%) high mild
  2 (4.00%) high severe
Queries/[LIGHT] Cram query all
                        time:   [594.48 ns 596.35 ns 598.44 ns]
Found 5 outliers among 50 measurements (10.00%)
  2 (4.00%) low severe
  1 (2.00%) low mild
  1 (2.00%) high mild
  1 (2.00%) high severe
Queries/[LIGHT] Cram query specific
                        time:   [598.58 ns 603.78 ns 609.88 ns]
Found 4 outliers among 50 measurements (8.00%)
[...]

PR #352, opened 2024-12-14
fix issue #248 - no stdout pollution on GH Actions anymore

sassman:
- this fixes issue #248
- on gitub actions, [a change on perf](https://lore.kernel.org/lkml/tip-d7065adb9b4f3384c2615f0a3dbdb6c3aae1eb18@git.kernel.org/T/) leads to a polluted stdout and no perf.data file written
- in cases where the user does not provide a `-o perf.data` argument as part of the custom cmd, the `-o perf.data` option is automatically always added

My first approach to introduce a tempfile caused permission issues when using the `--root` option. Hence aiming now for the most simple solution to the problem.


sassman:
@djc WDYT, could this be ok?

djc:
Fixed via #353, thanks for your contribution!

PR #355, opened 2024-12-19
Suppress output of `perf --help` during discovery

bobrik:
Without this the output is printed to stderr, where it's just confusing.

We already do this for dtrace.


djc:
Thanks!

PR #356, opened 2024-12-19
feat: support profiling of multiple PID

stormshield-guillaumed:
`perf` support passing multiple PID  with the `-p` flag with a comma separated list. For the flamegraph binary, this PR allows you to pass multiple PID, either with multiple `-p` flag, or a single `-p` flag with a comma separated list to mimic the perf CLI.

I tested my changes on Linux only. I checked the dtrace man page on mac and it seems to support multiple PID by passing the `-p` flag multiple times.

### DIFF
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -9,9 +9,9 @@ use flamegraph::Workload;
 #[derive(Debug, Parser)]
 #[clap(version)]
 struct Opt {
-    /// Profile a running process by pid
-    #[clap(short, long)]
-    pid: Option<u32>,
+    /// Profile a running process by pid (comma separated list)
+    #[clap(short, long, value_delimiter(','))]
+    pid: Vec<u32>,
 
     /// Generate shell completions for the given shell.
     #[clap(long, value_name = "SHELL", exclusive(true))]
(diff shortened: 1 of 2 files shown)


src/bin/flamegraph.rs
```diff
@@ -9,9 +9,9 @@ use flamegraph::Workload;
 #[derive(Debug, Parser)]
 #[clap(version)]
 struct Opt {
-    /// Profile a running process by pid
-    #[clap(short, long)]
-    pid: Option<u32>,
+    /// Profile a running process by pid (comma separated list)
+    #[clap(short, long, value_delimiter(','))]
+    pid: Option<Vec<u32>>,
```
djc, line 14: Can we avoid the `Option` wrapper by using the default for that case?

src/lib.rs
```diff
@@ -100,8 +100,10 @@ mod arch {
                 command.args(&c);
             }
             Workload::Pid(p) => {
+                let p = p.into_iter().map(|p| p.to_string()).collect::<Vec<_>>();
```
djc, line 103: Should be able to avoid the intermediate allocations here.

djc (approved):
Nice! Just a few nits...

src/bin/flamegraph.rs
stormshield-guillaumed: Behaviour of `Option<Vec<T>>` and `Vec<T>` is the same for clap, so I just removed the `Option`

src/lib.rs
stormshield-guillaumed: Not a big fan of the new solution, but the `join` method is only for slices of `Borrow<str>`, so I implemented it manually to make it one argument for perf.

djc:
Thanks!

stormshield-guillaumed:
It seems I broke the CI for macOS with my last change. Do I open a new PR ?

djc:
I was a bit too fast in merging it, but already fixed it in #357.

PR #363, opened 2025-02-10
Run stacks through rustc_demangle

### LINKED ISSUE #362: Symbols are not demangled properly with `-Csymbol-mangling-version=v0`
See: https://doc.rust-lang.org/rustc/symbol-mangling/v0.html

With default demangling:

Image

With `RUSTFLAGS=-Csymbol-mangling-version=v0`:

Image

That's on macOS, so dtrace based profiling.

There's a crate that supports better demangling: https://github.com/luser/rustfilt

bobrik:
This enables proper demangling of `RUSTFLAGS=-Csymbol-mangling-version=v0`, especially on macOS where the binary name is included in the frames.

Closes #362.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -515,6 +515,7 @@ dependencies = [
  "indicatif",
  "inferno",
  "opener",
+ "rustc-demangle",
  "shlex",
  "signal-hook",
 ]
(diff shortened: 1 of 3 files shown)


bobrik:
I tested on Linux and macOS with and without `RUSTFLAGS=-Csymbol-mangling-version=v0` and it works as expected.

djc (approved):
LGTM, thanks!

PR #364, opened 2025-02-17
Update `addr2line` setup

QuarticCat:
Gimli `addr2line` now can be installed directly via `cargo install`. It will automatically persist and take precedence over system's `addr2line`. (Unless you haven't setup your cargo environment, in which case you cannot `cargo install`.)


djc (approved):

djc:
Thanks!

PR #370, opened 2025-03-25
Make `initial_command()` return `Result`

ldm0:
addresses https://github.com/flamegraph-rs/flamegraph/pull/368#discussion_r2010046105

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,7 @@ use inferno::collapse::dtrace::{Folder, Options as CollapseOptions};
 #[cfg(unix)]
 use signal_hook::consts::{SIGINT, SIGTERM};
 
-use anyhow::{anyhow, Context};
+use anyhow::{anyhow, bail, Context};
 use clap::{
     builder::{PossibleValuesParser, TypedValueParser},
     Args,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -261,17 +260,19 @@ mod arch {
 
                         let trace = match blondie::trace_command(command_builder, false) {
                             Err(err) => {
-                                eprintln!("{}: {:?}", BLONDIE_ERROR, err);
-                                exit(1);
+                                bail!("{}: {:?}", BLONDIE_ERROR, err);
```
djc, line 263: Want to inline the format arguments while you're at it?

djc (approved):
Looks great, thanks!

src/lib.rs
ldm0: Done

src/lib.rs
```diff
@@ -259,19 +258,17 @@ mod arch {
...
 
-                        let trace = match blondie::trace_command(command_builder, false) {
-                            Err(err) => {
-                                eprintln!("{}: {:?}", BLONDIE_ERROR, err);
-                                exit(1);
-                            }
-                            Ok(trace) => trace,
-                        };
+                        let trace = blondie::trace_command(command_builder, false)
+                            .map_err(|err| anyhow!("{}: {:?}", BLONDIE_ERROR, err))?;
```
ldm0, line 262: `blondie::Error` doesn't `impl std::error::Error`, so we have to use `map_err + Debug` rather than `context`.

ldm0:
Fixed

djc (approved):

PR #376, opened 2025-05-02
doc: Remove mention of sudo/root from macOS installation section

corneliusroemer:
Root/sudo is no longer required on macOS since https://github.com/flamegraph-rs/flamegraph/pull/374 was merged.

Hence, the section of the README that explained why that was necessary can be removed.


djc:
Thanks!

PR #404, opened 2025-10-17
Ubuntu aarch: Add installation instructions and workaround

matthewsanetra:
I managed to get it working on Ubuntu aarch, added a workaround I've found.


djc (approved):
Thanks!

PR #405, opened 2025-10-19
Set blondie backend to default.

meipeter:


src/lib.rs
```diff
@@ -336,14 +336,14 @@ mod arch {
                 #[cfg(target_os = "windows")]
                 {
                     let mut help_test = crate::arch::base_dtrace_command(None);
-
+                    let if_dtrace_enabled = std::env::var("DTRACE").is_ok();
```
djc: This layers a bunch of stuff on top which is not the best approach. Should probably rework the `initial_command()` function to do something like this:

```rust
match (workload, env::var("DTRACE")) {
     #[cfg(target_os = "windows")]
     (Workload::Command(c), Err(_)) => // do blondie
     #[cfg(not(target_os = "windows"))]
     (Workload::Command(c), Err(_)) => // dtrace on path fallback
     (Workload::Command(c), Ok(dtrace) => // do dtrace
     _ => // do dtrace
}
```

PR #410, opened 2025-11-04
Add --compression-level/-z option for compressed perf record

alexheretic:
New option for recording compressed trace data. This can make a significant difference to the perf.data size.

```
  -z, --compression-level [<COMPRESSION_LEVEL>]
          Produce compressed trace using the specified level. perf only
```

## Example using (fastest) compression level `-z`
```
[ perf record: Woken up 11606 times to write data ]
[ perf record: Captured and wrote 96.617 MB perf.data, compressed (original 2950.013 MB, ratio is 30.572) ] 
```
What would have been ~3GB perf.data is ~100MB compressed without any noticeable overhead :slightly_smiling_face: 

## Alternatives
It is possible to do this already with `--cmd "record -F 997 --call-graph dwarf,64000 -g -z"` but this isn't as nice, and compression is something I want to use pretty much always.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,6 +38,7 @@ pub enum Workload {
 
 #[cfg(target_os = "linux")]
 mod arch {
+    use std::fmt::Write;
     use std::time::Duration;
 
     use indicatif::{ProgressBar, ProgressStyle};
@@ -51,6 +52,7 @@ mod arch {
         workload: Workload,
         sudo: Option<Option<&str>>,
         freq: u32,
+        compress_level: Option<u8>,
         custom_cmd: Option<String>,
         verbose: bool,
         ignore_status: bool,
@@ -72,7 +74,13 @@ mod arch {
         };
         let mut command = sudo_command(&perf, sudo);
 
-        let args = custom_cmd.unwrap_or(format!("record -F {freq} --call-graph dwarf,64000 -g"));
+        let args = custom_cmd.unwrap_or_else(|| {
+            let mut args = format!("record -F {freq} --call-graph dwarf,64000 -g");
+            if let Some(z) = compress_level {
+                _ = write!(args, " -z={z}");
+            }
+            args
+        });
 
         let mut perf_output = None;
         let mut args = args.split_whitespace();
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -72,7 +73,14 @@ mod arch {
         };
         let mut command = sudo_command(&perf, sudo);
 
-        let args = custom_cmd.unwrap_or(format!("record -F {freq} --call-graph dwarf,64000 -g"));
+        let args = custom_cmd.unwrap_or_else(|| {
+            use std::fmt::Write;
```
djc: Nit: import `fmt::Write` at the top of the module, please.

src/bin/cargo-flamegraph.rs
```diff
@@ -236,7 +236,7 @@ fn workload(opt: &Opt, artifacts: &[Artifact]) -> anyhow::Result<Vec<String>> {
             a.executable
                 .as_deref()
                 .filter(|_| {
-                    a.target.name == *target && a.target.kind.iter().any(|k| kind.contains(&k))
+                    a.target.name == *target && a.target.kind.iter().any(|k| kind.contains(k))
```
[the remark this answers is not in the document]
alexheretic: Yep clippy::needless_borrow

src/lib.rs
```diff
@@ -670,6 +686,16 @@ impl Options {
...
+
+    #[cfg(target_os = "linux")]
+    fn compression_level(&self) -> Option<u8> {
+        self.compression_level
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    fn compression_level(&self) -> Option<u8> {
+        None
+    }
```
djc: Would prefer to keep this in the caller. I think this should work?

```rust
#[cfg(target_os = "linux")]
let compression = opts.compression_level;
#[cfg(not(target_os = "linux"))]
let compression = None;
```

djc (approved):
Please split and squash so that we end up with two commits:

- Apply clippy suggestions
- Add compression level option for compressed perf records on Linux

src/lib.rs
alexheretic: sure

alexheretic:
> Please split and squash so that we end up with two commits

Done.

djc:
Released as 0.6.10.

PR #424, opened 2026-01-03
quick-xml demangling

cjgatto:
Encountered same "invalid xml from xctrace" detailed at https://github.com/flamegraph-rs/flamegraph/issues/421

Stole the fix https://github.com/flamegraph-rs/flamegraph/commit/f82bc924e9e614d0b15203f01b9d2770ea74c373

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,8 @@
 use std::{
+    borrow::Cow,
     env,
     fs::File,
-    io::{BufReader, BufWriter, Cursor, Read, Write},
+    io::{BufRead, BufReader, BufWriter, Cursor, Error, ErrorKind, Read, Write},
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -490,6 +491,57 @@ fn print_command(cmd: &Command, verbose: bool) {
     }
 }
 
+fn demangle_stream<R: BufRead, W: Write>(input: &mut R, output: &mut W) -> std::io::Result<()> {
```
djc: We should only do this on macOS when the xctrace backend is used.

djc (commented):
Here's some early feedback. I'll likely have more when this is addressed.

djc:
Merged via

- #435 

Thanks!

PR #430, opened 2026-01-29
Consider rlib targets in addition to lib targets

### LINKED ISSUE #422: Failes when `[lib]` section included in Cargo.toml
adding
```toml
[lib]
crate-type = ["cdylib", "rlib", "staticlib"]
```
to the Cargo.toml of a crate inside my workspace makes flamegraph unable to recognize the crate as a valid target, it just gives the error
```
Error: crate has no automatically selectable target:
```

divergentdave:
This changes target filtering in cargo-flamegraph to consider `TargetKind::Rlib` in addition to `TargetKind::Lib`. If the `crate-type` metadata has been customized, such that it no longer contains the default of `lib`, this affects the `TargetKind` seen in the list of targets produced by `cargo metadata`. The `lib` crate type acts as an alias for a different library type, depending on context. Thus, we should consider `rlib` targets when trying to run unit tests, just as we consider `lib`. The `--unit-test-kind` CLI option still has possible values of `bin` and `lib`, but the latter now checks for both kinds of library targets. This closes #422. I have tested this on crates with a `crate-type` that includes `lib`, `rlib`, and both.

### DIFF
--- a/src/bin/cargo-flamegraph.rs
+++ b/src/bin/cargo-flamegraph.rs
@@ -150,7 +150,7 @@ fn build(opt: &Opt, kind: Vec<TargetKind>) -> anyhow::Result<Vec<Artifact>> {
     if let Some(Some(ref unit_test)) = opt.unit_test {
         match kind
             .iter()
-            .any(|k| [TargetKind::Lib, TargetKind::RLib].contains(k))
+            .any(|k| matches!(k, TargetKind::Lib | TargetKind::RLib))
         {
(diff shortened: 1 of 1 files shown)


djc (approved):
LGTM! Curious what you're benchmarking...

divergentdave:
This came up with [zk-cred-longfellow](https://github.com/abetterinternet/zk-cred-longfellow/). We customized `crate-type` because we're targeting WASM.

djc (approved):

PR #444, opened 2026-04-15
Update in readme info on new default linker and old debian releases

mladedav:
- `lld` linker is now the default so I tried to make it clearer that people have to add the configuration for flamegraphs to work proprely without explicitly setting the linker.
- `bullseye` is not stable, it's even past its EOL now and in another 5 months it will be past its extended EOL. Personally, I would probably just remove the note altogether, but I didn't want to do that on my own since people could conceivably still use it so it you want me to, I'll delete it, otherwise I'm just removing the stable/testing notes about bullseye and bookworm.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1409,9 +1409,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.10"
+version = "0.103.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


djc (approved):
Good stuff, thanks!

djc:
> * Personally, I would probably just remove the note altogether, but I didn't want to do that on my own since people could conceivably still use it so it you want me to, I'll delete it, otherwise I'm just removing the stable/testing notes about bullseye and bookworm.

Happy for any further updates here that you think make sense!

PR #446, opened 2026-04-20
fix(macos): drop Event::Empty arm in demangle_stream that corrupted x…

### LINKED ISSUE #445: Read xml event failed: IllFormed
flamegraph-flamegraph 0.6.12
https://github.com/flamegraph-rs/flamegraph/issues/421 fixed, But I still get (on macos) 
```
Target app exited, ending recording...
Recording completed. Saving output file...
Output file saved as: cargo-flamegraph.trace
Error: unable to collapse generated profile data

Caused by:
    Read xml event failed: IllFormed(MismatchedEndTag { expected: "frame", found: "backtrace" })
```

boguscoder:
The Event::Empty arm (self-closing tags like <binary .../>) was matched to 'start' and then written as Event::Start without a corresponding closing tag, producing malformed XML that broke inferno's xctrace collapser with IllFormed(MismatchedEndTag).

This broke profiling of C++ binaries on macOS.

This should fix https://github.com/flamegraph-rs/flamegraph/issues/445

Ive tested this with c++ and rust binaries but please lmk if more coverage would help

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -506,7 +506,6 @@ fn demangle_stream<R: BufRead, W: Write>(input: &mut R, output: &mut W) -> std::
         let start = match reader.read_event_into(&mut buf) {
             Ok(Event::Eof) => break,
             Ok(Event::Start(start)) => start,
-            Ok(Event::Empty(start)) => start,
             Ok(el) => {
                 writer.write_event(el)?;
                 continue;


djc (approved):

djc:
Sounds good, thanks. How hard would it be to have a (smallish) sample XML file in the repo to exercise this code against?

boguscoder:
Should be easy, i'll provide file soon but whether it needs some CI hooking into is a mystery to me

djc:
> Should be easy, i'll provide file soon but whether it needs some CI hooking into is a mystery to me

If you can just submit a PR with a unit test that would be great.

nixpulvis:
Thanks, seems to have fixed things for me too.

PR #456, opened 2026-06-26
Automatically add `--no-rosegment` flag

itsjunetime:
I regularly run my projects with various different profiles and configurations, and regularly forget to include the `--no-rosegment` flag when I try to run perf/flamegraph stuff. This causes some annoyance and confusion when I get the flamegraph and it doesn't show me any useful information, so I figured it would be nice to try to detect when it's needed and include it automatically.

To do so, this PR:
1. Checks the default cargo toolchain's rust version
2. Gets rustflags from the `RUSTFLAGS` env var and the target's rustflags settings within `cargo/config.toml` using `cargo config get` (on nightly; if the user doesn't have the nightly toolchain installed, it just silently doesn't check the flags from that source)
3. Goes through them and detects:
    a. If they're using `gold`, which doesn't require this flag (according to the README)
    b. If they are explicitly using a linker which requires it
    c. If they've already specified the `--no-rosegment` flag

If ((they're using rust >= 1.90 && not explicitly using gold) || using a linker that needs this flag) && not already specifying the no-rosegment flag, then we add it to the `RUSTFLAGS` env var for cargo compilation.

I've also added some tests since this has some potentially finnicky parsing behavior, and adding configuration to CI to run those tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       with:
         toolchain: stable
         components: rustfmt, clippy
-    - uses: dtolnay/rust-toolchain@nightly
+    - run: rustup toolchain install nightly # we need nightly for some tests which try to detect rustflags
     - run: cargo clippy --all-targets --all-features -- -D warnings
     - run: cargo fmt -- --check
     - run: cargo test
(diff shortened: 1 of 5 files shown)


itsjunetime:
Hmm, yeah I made some changes to CI and I guess I didn't understand how to use the rust-toolchain action correctly. Gotta poke around to fix it.

djc (commented):
This seems to add a lot of specifically targeted complexity that I don't really feel confident reviewing, to the point that I'm not sure it's a good fit for the flamegraph tooling.

Issue #462, opened 2026-07-13
Demangled calls containing semicolons split across multiple lines on MacOS


CraftSpider:
I'm not certain if this error is MacOS specific for sure, but using recent `cargo flamegraph` with xctrace backend, functions with array types in them are being split across multiple lines on the semicolon. I've attached an example svg, generated from the following code:

```
//! Run with `cargo flamegraph --ignore-status`
use std::hint::black_box;
use std::time;
use std::time::Duration;

fn foo<T: Default>() -> T {
    let start = time::SystemTime::now();
    while (time::SystemTime::now().duration_since(start).unwrap()) < Duration::from_secs(1) {
        black_box(())
    }
    black_box(T::default())
}

fn main() {
    for _ in 0..10 {
        println!("{:?}", foo::<[u8; 2]>());
    }
}
```

<img width="1200" height="342" alt="Image" src="https://github.com/user-attachments/assets/bc4d035f-7852-48ae-8ca8-f317f0d32e51" />

The `--ignore-status` is needed for an unrelated issue. Originally noticed while profiling something with quite a few array types in the demangled type, leading to comedically tall stacks.

djc:
flamegraph is mostly getting passive maintenance (have you tried samply?), so I won't be able to spend much time on this -- but happy to review a PR if you figure out the problem.

PR #464, opened 2026-08-05
fix: keep semicolons inside xctrace symbols

tomatotomata:
## Summary

xctrace folded stacks use semicolons to separate frames. Rust demangling can introduce semicolons inside array types such as `[u8; 2]`, so those symbols were being split into multiple frames.

I was thinking of applying the same normalization already used by inferno's DTrace collapser: replace semicolons inside the demangled symbol with colons before the XML reaches the xctrace collapser. The change is limited to the macOS demangling path and includes a regression test for an array type.

## Validation

- `cargo fmt -- --check`
- `cargo test --all-targets`
- `git diff --check`

The test suite passed on Windows. I could not run the macOS-only test locally because the Apple target is not installed here.

This follows the maintainer's invitation in #462. Let me know what you think.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -496,6 +496,13 @@ fn print_command(cmd: &Command, verbose: bool) {
 }
 
 #[cfg(target_os = "macos")]
+fn normalize_demangled_symbol(symbol: String) -> String {
+    // Folded stacks use semicolons as frame separators. Keep semicolons that
+    // occur inside Rust array types from being interpreted as extra frames.
+    symbol.replace(';', ":")
+}
+
+#[cfg(target_os = "macos")]
 fn demangle_stream<R: BufRead, W: Write>(input: &mut R, output: &mut W) -> std::io::Result<()> {
     let mut reader = quick_xml::Reader::from_reader(input);
     let mut writer = quick_xml::Writer::new(output);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -495,6 +495,13 @@ fn print_command(cmd: &Command, verbose: bool) {
     }
 }
 
+#[cfg(target_os = "macos")]
+fn normalize_demangled_symbol(symbol: String) -> String {
```
djc: Just inline this.

src/lib.rs
```diff
@@ -537,6 +544,19 @@ fn demangle_stream<R: BufRead, W: Write>(input: &mut R, output: &mut W) -> std::
     Ok(())
 }
 
+#[cfg(all(test, target_os = "macos"))]
+mod tests {
```
djc: Any tests modules should be at the bottom of the file.

src/lib.rs
tomatotomata: Inlined the semicolon normalization at the demangling site in 48b191e.

src/lib.rs
tomatotomata: Moved the test module to the bottom of src/lib.rs. I also changed it to exercise demangle_stream with a real Rust v0 symbol, so it covers the XML boundary instead of the removed helper.

djc (approved):

djc:
Thanks!

